PR #131, opened 2015-01-21
Fix for valid html documents which would always mark tests as skipped.

peschee:
Valid documents return an empty array. Empty arrays always triggered skipping of tests, as opposed to running through rendering tests valid.


peschee:
Just saw that there's already a PR similar to this #114. It seems that #114 fixes this (as well as some additional issues) so I guess that one should be merged rather than this?

lsmith77:
ok .. I merged #114

peschee:
Thanks @lsmith77

Issue #132, opened 2015-01-23
Make this bundle usable for browser end-to-end tests


thomas-koch-comsolit:
It seems, this bundle is mostly used for tests with PHPUnit and WebTestCase. We want to also do tests in a browser (e.g. via [1][2]). This means we need to include test setup information in the HTTP requests (via Query params?):

- What fixtures to load
- What user to login as

e.g.

```
localhost:8000/someroute?_login=admin&_fixtures=Bamarni\MainBundle\DataFixtures\ORM\LoadData,Me\MyBundle\DataFixtures\ORM\LoadData
```

And we need a listener to do this on KernelEvents::REQUEST(?).

[1] http://angular.github.io/protractor
[2] https://github.com/minkphp

I'll work on this next week unless somebody claims this would be a silly idea. Would you take a PR for this?

lsmith77:
Its not quite clear to me what you want to do ..

thomas-koch-comsolit:
Maybe you'd like to have a look at https://github.com/comsolit/LiipFunctionalTestBundle/tree/wip_make_e2e_tests_possible

I've extracted all Database related stuff in a standalone class https://github.com/comsolit/LiipFunctionalTestBundle/blob/wip_make_e2e_tests_possible/Database/TestDatabasePreparator.php

The TestDatabasePreparator is used from the https://github.com/comsolit/LiipFunctionalTestBundle/blob/wip_make_e2e_tests_possible/EventListener/PrepareDatabaseListener.php

So in Jenkins I start a PHP server in the test environment and can issue requests against the application and for each request the DB is initialized with a set of fixtures that I can select with the '_fixtures' GET parameter.

PR #136, opened 2015-01-28
Make e2e tests possible

ghost:
This branch is a prototype to solve #132 
The code is clean enough to be merged but should be made more configurable and get some documentation. But I'd like to get your feedback first.

Now you can run your app in the test environment and use to new GET parameters:
- _fixtures=/Some/Fixture/Class,/More/Fixtures/Class
- _login=old-joe-42

This is a snippet from my config_test.yml that uses the new listeners:

```
security:
    firewalls:
        secured_area:
            http_basic: ~

liip_functional_test:
    cache_sqlite_db: true

doctrine:
    dbal:
        default_connection: default
        connections:
            default:
                driver: pdo_sqlite
                path: "%kernel.cache_dir%/test.db"

services:
    liip_functional_test.prepare_database:
        class: Liip\FunctionalTestBundle\EventListener\PrepareDatabaseListener
        arguments:
            - "@service_container"
        tags:
            - { name: kernel.event_subscriber }
    liip_functional_test.login_user:
        class: Liip\FunctionalTestBundle\EventListener\LoginTestUserListener
        tags:
            - { name: kernel.event_subscriber }
```

### DIFF
--- a/Database/TestDatabaseCache.php
+++ b/Database/TestDatabaseCache.php
@@ -12,6 +12,7 @@
 namespace Liip\FunctionalTestBundle\Database;
 
 use Symfony\Component\DependencyInjection\Container;
+use Doctrine\Common\DataFixtures\Executor\AbstractExecutor;
 
 class TestDatabaseCache
 {
@@ -20,6 +21,30 @@ class TestDatabaseCache
     public function __construct(Container $container)
     {
         $this->container = $container;
+    }
+
+    public function getCachedExecutor(TestDatabasePreparator $dbPreparator, array $classNames)
+    {
+        $backup = $this->buildCacheFilePath($dbPreparator->getMetaDatas(), $classNames);
+        if ($this->isBackupUpToDate($classNames, $backup)) {
+            $om = $dbPreparator->getObjectManager();
+            $om->flush();
+            $om->clear();
+
+            $executor = $dbPreparator->getExecutorWithReferenceRepository();
+            $executor->getReferenceRepository()->load($backup);
+
+            copy($backup, $this->getSQLiteName($om->getConnection()->getParams()));
+
+            return $executor;
+        }
+    }
+
+    public function storeToCache(AbstractExecutor $executor, TestDatabasePreparator $dbPreparator, array $classNames)
+    {
+        $backup = $this->buildCacheFilePath($dbPreparator->getMetaDatas(), $classNames);
+        $executor->getReferenceRepository()->save($backup);
+        copy($this->getSQLiteName($dbPreparator->getObjectManager()->getConnection()->getParams()), $backup);
     }
 
     /**
@@ -60,7 +85,7 @@ class TestDatabaseCache
      * @return bool TRUE if the backup was made since the modifications to the
      *         fixtures; FALSE otherwise
      */
-    public function isBackupUpToDate(array $classNames, $backup)
+    private function isBackupUpToDate(array $classNames, $backup)
     {
         if(!file_exists($backup) || !file_exists($backup.'.ser')) {
             return false;
@@ -85,7 +110,7 @@ class TestDatabaseCache
      * @param array $classNames
      * @return string full path to the database cache file
      */
-    public function buildCacheFilePath(array $metadatas, array $classNames)
+    private function buildCacheFilePath(array $metadatas, array $classNames)
     {
         return $this->container->getParameter('kernel.cache_dir') . '/test_' . md5(serialize($metadatas) . serialize($classNames)) . '.db';
     }
--- a/Database/TestDatabaseCache.php
+++ b/Database/TestDatabaseCache.php
@@ -1,0 +1,106 @@
+<?php
+
+/*
+ * This file is part of the Liip/FunctionalTestBundle
+ *
+ * (c) Lukas Kahwe Smith <smith@pooteeweet.org>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Liip\FunctionalTestBundle\Database;
+
+use Symfony\Component\DependencyInjection\Container;
+
+class TestDatabaseCache
+{
+    private $container;
+
+    public function __construct(Container $container)
+    {
+        $this->container = $container;
+    }
+
+    /**
+     * This function finds the time when the data blocks of a class definition
+     * file were being written to, that is, the time when the content of the
+     * file was changed.
+     *
+     * @param string $class
+     *            The fully qualified class name of the fixture class to
+     *            check modification date on.
+     *
+     * @return \DateTime|null
+     */
+    protected function getClassLastModified($class)
+    {
+        $lastModifiedDateTime = null;
+
+        $reflClass = new \ReflectionClass($class);
+        $classFileName = $reflClass->getFileName();
+
+        if (file_exists($classFileName)) {
+            $lastModifiedDateTime = new \DateTime();
+            $lastModifiedDateTime->setTimestamp(filemtime($classFileName));
+        }
+
+        return $lastModifiedDateTime;
+    }
+
+    /**
+     * Determine if the Fixtures that define a database backup have been
+     * modified since the backup was made.
+     *
+     * @param array $classNames
+     *            The fixture classnames to check
+     * @param string $backup
+     *            The fixture backup SQLite database file path
+     *
+     * @return bool TRUE if the backup was made since the modifications to the
+     *         fixtures; FALSE otherwise
+     */
+    public function isBackupUpToDate(array $classNames, $backup)
+    {
+        if(!file_exists($backup) || !file_exists($backup.'.ser')) {
+            return false;
+        }
+
+        $backupLastModifiedDateTime = new \DateTime();
+        $backupLastModifiedDateTime->setTimestamp(filemtime($backup));
+
+        foreach ($classNames as &$className) {
+            $fixtureLastModifiedDateTime = $this->getClassLastModified($className);
+            if ($backupLastModifiedDateTime < $fixtureLastModifiedDateTime) {
+                return false;
+            }
+        }
+
+        return true;
+    }
+
+    /**
+     *
+     * @param array $metadatas
+     * @param array $classNames
+     * @return string full path to the database cache file
+     */
+    public function buildCacheFilePath(array $metadatas, array $classNames)
+    {
+        return $this->container->getParameter('kernel.cache_dir') . '/test_' . md5(serialize($metadatas) . serialize($classNames)) . '.db';
+    }
+
+    /**
+     * Is caching of SQLite Database instances enabled?
+     *
+     * Caching can be enabled in app/config/config_test.yml:
+     * ``` yaml
+     * liip_functional_test:
+     *     cache_sqlite_db: true
+     * ```
+     */
+    public function isCacheEnabled()
+    {
+        return $this->container->getParameter('liip_functional_test.cache_sqlite_db');
+    }
+}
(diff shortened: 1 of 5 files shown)


EventListener/PrepareDatabaseListener.php
```diff
@@ -0,0 +1,68 @@
...
+            return;
+        }
+
+        $classNames = $this->getFixturesClassNames($event->getRequest());
+        if (is_null($classNames)) {
+            return;
+        }
+
+        // @TODO Make the registry service name configurable
+        $registry = $this->container->get('doctrine');
```
lsmith77: this seems like an important thing to solve before we can merge

EventListener/PrepareDatabaseListener.php
```diff
@@ -0,0 +1,68 @@
...
+
+        $dbPreparator = new TestDatabasePreparator($this->container, $registry);
+        $dbPreparator->loadFixtures($classNames);
+    }
+
+    private function getFixturesClassNames(Request $request)
+    {
+        // @TODO Make the query parameter name configurable
+        $fixturesParam = $request->query->get('_fixtures');
```
lsmith77: this to me is a nice to have

EventListener/LoginTestUserListener.php
```diff
@@ -0,0 +1,38 @@
...
+class LoginTestUserListener implements EventSubscriberInterface
+{
+    const FAKEPASS = 'fakePass_42!';
+
+    public function onKernelRequest(GetResponseEvent $event)
+    {
+        $request = $event->getRequest();
+
+        // @TODO make the query parameter name configurable
+        $login = $request->query->get('_login');
```
lsmith77: another nice to have

lsmith77:
This looks like a very good change to me.
The services you show in your above config should imho be provided by the bundle but only enabled via Bundle config.

ie.
```
liip_functional_test:
    cache_sqlite_db: true
    prepare_db_listener: true
    login_user_listener: true
```

And then the docs should also be updated.

Now the changes themselves are currently not BC compatible. Either they need to be made BC compatible, or we need to bump to 2.0. If we do jump to 2.0, then I guess we should have another look what else we might want to change along the way.

thomas-koch-comsolit:
Thank you for the review. I don't see, where I broke BC? I added new features that do lack configurability, but I hope I didn't break BC.

I'm afraid that my boss (or client) won't let me work on this branch again for the next few weeks. Could Liip be so kind to do the remaining changes, please?

It would also be better done by Liip, because you might have more projects to test against, more Symfony versions and PHCR, which we don't have.

lsmith77:
ok .. I will see.
the BC break is just several protected methods in WebTestCase were removed from what I can tell.

thomas-koch-comsolit:
ping? Please don't let this branch die... :-(

Database/TestDatabaseCache.php
```diff
@@ -0,0 +1,152 @@
...
+class TestDatabaseCache
+{
+    private $container;
+
+    public function __construct(Container $container)
+    {
+        $this->container = $container;
+    }
+
+    public function getCachedExecutor(TestDatabasePreparator $dbPreparator, array $classNames)
```
stof: phpdoc is needed for all these public methods

Database/TestDatabaseCache.php
```diff
@@ -0,0 +1,152 @@
...
+     * file were being written to, that is, the time when the content of the
+     * file was changed.
+     *
+     * @param string $class
+     *            The fully qualified class name of the fixture class to
+     *            check modification date on.
+     *
+     * @return \DateTime|null
+     */
+    protected function getClassLastModified($class)
```
stof: does it need to be an extension point ? if no, it should be private

Database/TestDatabaseCache.php
```diff
@@ -0,0 +1,152 @@
...
+
+    public function storeToCache(AbstractExecutor $executor, TestDatabasePreparator $dbPreparator, array $classNames)
+    {
+        $backup = $this->buildCacheFilePath($dbPreparator->getMetaDatas(), $classNames);
+        $executor->getReferenceRepository()->save($backup);
+        copy($this->getSQLiteName($dbPreparator->getObjectManager()->getConnection()->getParams()), $backup);
+    }
+
+    /**
+     * This function finds the time when the data blocks of a class definition
```
stof: the phpdoc should have a short description (which should be on 1 line)

Database/TestDatabaseCache.php
```diff
@@ -0,0 +1,152 @@
...
+        if (file_exists($classFileName)) {
+            $lastModifiedDateTime = new \DateTime();
+            $lastModifiedDateTime->setTimestamp(filemtime($classFileName));
+        }
+
+        return $lastModifiedDateTime;
+    }
+
+    /**
+     * Determine if the Fixtures that define a database backup have been
```
stof: should be ``whether`` instead of ``if``

Database/TestDatabaseCache.php
```diff
@@ -0,0 +1,152 @@
...
+    private function isBackupUpToDate(array $classNames, $backup)
+    {
+        if(!file_exists($backup) || !file_exists($backup.'.ser')) {
+            return false;
+        }
+
+        $backupLastModifiedDateTime = new \DateTime();
+        $backupLastModifiedDateTime->setTimestamp(filemtime($backup));
+
+        foreach ($classNames as &$className) {
```
stof: why using references here ? this is a bad idea, especially given you pass it to a function expecting a non-reference just after, causing a reference mismatch (see http://jpauli.github.io/2014/06/27/references-mismatch.html for the explanation)

Database/TestDatabaseCache.php
```diff
@@ -0,0 +1,152 @@
...
+            }
+        }
+
+        return true;
+    }
+
+    /**
+     *
+     * @param array $metadatas
+     * @param array $classNames
```
stof: ``string[]``

Database/TestDatabaseCache.php
```diff
@@ -0,0 +1,152 @@
...
+            throw new \InvalidArgumentException("Connection does not contain a 'path' or 'dbname' parameter and cannot be dropped.");
+        }
+
+        return $name;
+    }
+
+    /**
+     * Is caching of SQLite Database instances enabled?
+     *
+     * Caching can be enabled in app/config/config_test.yml:
```
stof: The explanation of how to enable it in the bundle config has nothing to do here. This class is not aware of the semantic config (and should not be).

Database/TestDatabaseCache.php
```diff
@@ -0,0 +1,152 @@
...
+     * ``` yaml
+     * liip_functional_test:
+     *     cache_sqlite_db: true
+     * ```
+     */
+    public function isCacheEnabled()
+    {
+        return $this->container->getParameter('liip_functional_test.cache_sqlite_db');
```
stof: instead of injecting the whole container to be able to retrieve 2 parameters from it, you should inject only these 2 scalars

Database/TestDatabasePreparator.php
```diff
@@ -0,0 +1,257 @@
...
+    /**
+     * @var array
+     */
+    static private $cachedMetadatas = [];
+
+    public function __construct(Container $container, ManagerRegistry $registry, $omName = null)
+    {
+        $this->container = $container;
+        $this->omName = $omName;
+        $this->om = $registry->getManager($omName);
```
stof: you should rather keep the registry and retrieve the object manager from it when needed, to be compatible with manager resetting

EventListener/LoginTestUserListener.php
```diff
@@ -0,0 +1,38 @@
...
+            [
+                'PHP_AUTH_USER' => $login,
+                'PHP_AUTH_PW'   => self::FAKEPASS,
+            ]
+        );
+    }
+
+    public static function getSubscribedEvents()
+    {
+        return [KernelEvents::REQUEST => ['onKernelRequest', 255]];
```
stof: the short array syntax is not compatible with the min PHP requirement, which is PHP 5.3

EventListener/LoginTestUserListener.php
```diff
@@ -0,0 +1,38 @@
+<?php
```
stof: the license header is missing

EventListener/LoginTestUserListener.php
```diff
@@ -0,0 +1,38 @@
...
+        $login = $request->query->get('_login');
+
+        if(is_null($login)) {
+            return;
+        }
+
+        $request->headers->add(
+            [
+                'PHP_AUTH_USER' => $login,
+                'PHP_AUTH_PW'   => self::FAKEPASS,
```
stof: this works only if your project uses HTTP basic

Test/WebTestCase.php
```diff
@@ -234,106 +194,19 @@ protected function isBackupUpToDate(array $classNames, $backup)
...
-        $executor->execute($loader->getFixtures(), true);
-
-        if (isset($name) && isset($backup)) {
-            $executor->getReferenceRepository()->save($backup);
-            copy($name, $backup);
-        }
+        return $dbPreparator->loadFixtures(
+            $classNames,
+            $purgeMode,
+            [$this, 'loadFixturesEventCallback']
```
stof: PHP 5.3 issue

Test/WebTestCase.php
```diff
@@ -234,106 +194,19 @@ protected function isBackupUpToDate(array $classNames, $backup)
...
+        );
+    }
 
-        return $executor;
+    public function loadFixturesEventCallback($methodName)
+    {
+        call_user_func([$this, $methodName]);
```
stof: ``$this->$methodName()`` will be faster than ``call_user_func``

Database/TestDatabaseCache.php
stof: btw, this method seems to be called only inside this class, so why is it public ?

Database/TestDatabaseCache.php
```diff
@@ -0,0 +1,152 @@
...
+            $executor = $dbPreparator->getExecutorWithReferenceRepository();
+            $executor->getReferenceRepository()->load($backup);
+
+            copy($backup, $this->getSQLiteName($om->getConnection()->getParams()));
+
+            return $executor;
+        }
+    }
+
+    public function storeToCache(AbstractExecutor $executor, TestDatabasePreparator $dbPreparator, array $classNames)
```
stof: same question here: why is it public ?

EventListener/PrepareDatabaseListener.php
```diff
@@ -0,0 +1,68 @@
...
+        }
+
+        $classNamesUnfiltered = explode(',', $fixturesParam);
+
+        return array_filter($classNamesUnfiltered);
+    }
+
+    public static function getSubscribedEvents()
+    {
+        return [KernelEvents::REQUEST => ['onKernelRequest', 255]];
```
stof: PHP 5.3 issue

nicolas-bastien:
@thomas-koch-comsolit I think your services should be loaded dynamicaly in DependancyInjection

As this bundle is instanciated in test / dev env only this would simplify the installation.

Maybe I can let you corrects the conflict and PR this change if you want

robertfausk:
Will there be any progress on this?

lsmith77:
nobody is working on it atm

robertfausk:
So what's the best way to start working on it again?
Afaik I'm not able to work directly on this PR.
So my suggestion is:
- [ ] create a new pull request with work of @ghost included
- [ ] resolve issues @stof mentioned
- [ ] resolve existing conflicts with current master branch

lsmith77:
sounds good!

alexislefebvre:
I'm closing this since fixtures loading has been moved to https://github.com/liip/LiipTestFixturesBundle

PR #139, opened 2015-02-08
WebTestCase: preFixtureRestore() method

kniziol:
Executed **before** restore fixtures. Required to do something before load the fixtures.

### DIFF
--- a/Test/WebTestCase.php
+++ b/Test/WebTestCase.php
@@ -30,6 +30,7 @@ use Symfony\Component\HttpFoundation\Session\Session;
 use Doctrine\Common\Persistence\ObjectManager;
 use Doctrine\Common\Persistence\ManagerRegistry;
 use Doctrine\Common\DataFixtures\DependentFixtureInterface;
+use Doctrine\Common\DataFixtures\Executor\AbstractExecutor;
 use Doctrine\Common\DataFixtures\ProxyReferenceRepository;
 
 use Doctrine\DBAL\Driver\PDOSqlite\Driver as SqliteDriver;
@@ -330,8 +331,13 @@ abstract class WebTestCase extends BaseWebTestCase
         $executor->execute($loader->getFixtures(), true);
 
         if (isset($name) && isset($backup)) {
+            $om = $executor->getObjectManager();
+            $this->preReferenceSave($om, $executor, $backup);
+
             $executor->getReferenceRepository()->save($backup);
             copy($name, $backup);
+
+            $this->postReferenceSave($om, $executor, $backup);
         }
 
         return $executor;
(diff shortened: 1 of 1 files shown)


lsmith77:
fine for me .. wondering if we should also add those parameters to the "post" method?

kniziol:
@lsmith77 It's good suggestionsə, I think

lsmith77:
can you add it to the PR?

kniziol:
@lsmith77 Yes. It's done.

kniziol:
I have created ```preReferenceSave()``` and ```postReferenceSave()``` methods

Test/WebTestCase.php
```diff
@@ -346,9 +353,51 @@ protected function postFixtureSetup()
...
+
+    /**
+     * Callback function to be executed before save of references
+     *
+     * @param ObjectManager $manager The object manager
+     * @param AbstractExecutor $executor Executor of the data fixtures
+     * @param string $backupFilePath Path of file used to backup the references of the data fixtures
+     * @return WebTestCase
      */
-    protected function postFixtureRestore()
```
lsmith77: actually that was a stupid suggestion by me .. because it will break BC, so can you please remove the new parameters to ``postFixtureRestore()``?

kniziol: OKey, I will remove those arguments, but... it will be good to access instance of the ```ObjectManager``` or ```ProxyReferenceRepository``` directly in ```postFixtureRestore()``` method. What do you think about it?

kniziol: @lsmith77 are you there? :]

lsmith77: sorry .. I will be off for 10 days and work as been getting ever so more stressful in the lead up time. so open source work kind of got the short end of the stick. I hope to wrap this up in the first days once I am back on the 15th.

lsmith77: ok I am back.

yeah it would be good but I would like to avoid BC breaks unless we do a bigger refactoring. Now I am open to such a bigger refactoring (see https://github.com/liip/LiipFunctionalTestBundle/pull/136/files) in case someone has time to finish it up. Until then it would be better to revert.

kniziol: I understand, but when refactoring will be finished?

lsmith77: I do not know, as you may have noticed this Bundle is not necessarily on the top of my agenda right now. The above PR is already quite far along. So if you have some time to work on it .. it could likely get merged soon.

kniziol: I don't have time :]
So, what finally do with arguments of the ```postFixtureRestore()``` method?

lsmith77: keep them as they were before this PR, ie. revert the change in this PR

kniziol: OKay. It's done.

Issue #145, opened 2015-03-17
loadFixtures should be able to DELETE/TRUNCATE a table with foreign keys


J-Ben87:
I have a bunch of tables, some of them with foreign keys.
When I try to reload my fixtures, I get the following error:

```
SQLSTATE[23000]: Integrity constraint violation: 1451 Cannot delete or update a parent row: a foreign key constraint fails
```
Is there a way to force deleting/truncating these tables anyway?

The solution I have implemented is to extend WebTestCase and override loadFixtures():

```
class MyWebTestCase extends WebTestCase
{
    /**
     * {@inheritdoc}
     */
    protected function loadFixtures(array $classNames, $omName = null, $registryName = 'doctrine', $purgeMode = null)
    {
        $connection = $this->getContainer()->get('doctrine')->getManager()->getConnection();

        $connection->query(sprintf('SET FOREIGN_KEY_CHECKS=0'));

        $result = parent::loadFixtures($classNames, $omName, $registryName, $purgeMode);

        $connection->query(sprintf('SET FOREIGN_KEY_CHECKS=1'));

        return $result;
    }
}
```
What do you think about it? If you are ok with that solution I can make a PR for you.

lsmith77:
this makes sense but should have a check if the DB is in fact MySQL

J-Ben87:
Ok, @nicolas-bastien PR fixed this, thanks

Issue #143, opened 2015-03-12
Question before PR - load yml file instead of fixture class


nicolas-bastien:
Hi,

We want to load our fixtures for testing directly with yml to be able to choose for each test which file we want to load.

Right now we extend your bundle to add this functionality directly in our project.
This functionality use Nelmoi/Alice to parse yml files

I think it can be a good idea to add it directly in LiipFunctionalTestBundle.

I can do the PR but before working on it, I want to check with you if this can be merged.

Here is an example : 

        $this->loadFixtureFiles([
            '@AppBundle/Tests/Ressources/DataFixtures/data/areaRange.yml',
            '@AppBundle/Tests/Ressources/DataFixtures/data/priceRange.yml',
        ]);

What do you think about it ?

lsmith77:
sounds good

nicolas-bastien:
done @lsmith77 maybe we can tag 1.1.1 before the refactoring PR ?

PR #146, opened 2015-03-20
#143 add loadFixtureFiles with Nelmio alice

nicolas-bastien:
Hi @lsmith77 ,

So here is my code. I added the foreign keys check as we need it on our project with the platform test as you suggest : https://github.com/liip/LiipFunctionalTestBundle/issues/145

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,13 +18,13 @@
         "php": ">=5.3.2",
         "symfony/framework-bundle": "~2.0",
         "symfony/browser-kit": "~2.0",
-        "doctrine/common": "2.*",
-        "nelmio/alice": "~1.7"
+        "doctrine/common": "2.*"
     },
     "suggest": {
         "doctrine/doctrine-fixtures-bundle": "Required when using the fixture loading functionality",
         "doctrine/orm": "Required when using the fixture loading functionality with an ORM and SQLite",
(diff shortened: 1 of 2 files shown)


lsmith77:
One thing that I am wondering is in light of this https://github.com/liip/LiipFunctionalTestBundle/pull/136 quite sensible refactoring to split out the fixture loading into a separate class, maybe we should do this with this feature already?

Or better yet, are you motivated to also wrap up the above noted PR?

nicolas-bastien:
When doing this PR I had the same idea of refactoring.

Maybe we can merge this first cause has this is not a big change.

then I'll help @thomas-koch-comsolit on is PR cause I think his refactoring needs a little more work.
Like @stof notes, documentation, services loading....

composer.json
```diff
@@ -18,7 +18,8 @@
         "php": ">=5.3.2",
         "symfony/framework-bundle": "~2.0",
         "symfony/browser-kit": "~2.0",
-        "doctrine/common": "2.*"
+        "doctrine/common": "2.*",
+        "nelmio/alice": "~1.7"
```
lsmith77: this should imho be an optional dependency. in this sense there should maybe then also be an explicit check in the code if the class exists before making an instance.

lsmith77:
yeah, its ok for me to merge it in (see my one comment).

nicolas-bastien:
@lsmith77 done

PR #147, opened 2015-03-21
fix type hints and doc comments

Ma27:


Test/WebTestCase.php
```diff
@@ -554,7 +555,7 @@ protected function getUrl($route, $params = array(), $absolute = false)
      *
      * @return void
      */
-    public function isSuccessful($response, $success = true, $type = 'text/html')
+    public function isSuccessful(Response $response, $success = true, $type = 'text/html')
```
lsmith77: since others might have extended this method it would be a BC break. then again most users likely didn't extend this method. still imho as we are looking towards a 2.0 release, this change should be in a separate PR that we will merge once we begin on 2.0

Ma27: well, I don't know the history of that bundle (which I do really love by the way), but the ApiDoc tells that a response object is requried and this method uses internally the methods of the symfony response object so I cannot understand why there's a BC break.

Ma27:
ok, so I should revert this change and open another PR for this change @lsmith77 ??

lsmith77:
@Ma27 yes

lsmith77:
sorry for not having gotten back earlier :-/

lsmith77:
master was cleaned up quite a bit now via styleci

Issue #149, opened 2015-03-31
Problem with Version 1.2.0 on php 5.3.10


jimtonic:
The minimum php version to use this bundle is 5.3.2 according to the composer.json.
However in Tests/WebTestCase on line 373 you made used the short array syntax, that isn't available before php 5.4
So, since I updated the LiipFunctionalTestBundle from 1.1.0 to 1.2.0 I get the following error:

```
Parse error: syntax error, unexpected '[' in /opt/jenkins/ferdi/jobs/Services - BPP build/workspace/vendor/liip/functional-test-bundle/Test/WebTestCase.php on line 373
```

lsmith77:
can you submit a PR to fix it?

jimtonic:
sure, i have submitted a PR.

lsmith77:
closed via https://github.com/liip/LiipFunctionalTestBundle/releases/tag/1.2.1

PR #151, opened 2015-04-01
fixed issue #149 - short array syntax is not available on php 5.3

jimtonic:
here we go...


lsmith77:
thx! see https://github.com/liip/LiipFunctionalTestBundle/releases/tag/1.2.1

PR #156, opened 2015-05-06
fix #155

### LINKED ISSUE #155: loadFixtures backup gets invalidated
$metadatas sometimes changes order of classes, although the metadata content is actually the same.

Particularly when the app is run the first time, with cache cleared - one class from our app just jumps a few places in the array.

Here is the $metadatas array, with md5 hashes of serialized values: https://www.diffchecker.com/uawo873x

kokoon:
fixes #155

### DIFF
--- a/Test/WebTestCase.php
+++ b/Test/WebTestCase.php
@@ -288,6 +288,7 @@ abstract class WebTestCase extends BaseWebTestCase
 
                 if (!isset(self::$cachedMetadatas[$omName])) {
                     self::$cachedMetadatas[$omName] = $om->getMetadataFactory()->getAllMetadata();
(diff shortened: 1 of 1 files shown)


lsmith77:
so with this you are ensuring that the order is always the same by forcing the order?

kokoon:
That's the idea, yes. If only the metadatas order changes, I would still consider the fixtures backup valid, since the metadata contents aren't changed.

lsmith77:
ok .. I trust you tested it with your setup .. I currently do not have a good project to test this against.

kokoon:
I have tested it, fixes my problem and I couldn't think of any cases that
would break.
Worst case scenario is, next time somebody that has the patch runs the
tests with the same metadatas they have recently, the backup gets
invalidated and fixtures will be set up again. Just like after clearing the
cache.

PR #157, opened 2015-05-24
fix doc url

lazyants:


lsmith77:
are you using the validator integration? i am not sure if it still works as I am not sure how many people are using it.

lazyants:
no, I just noticed that the link has changed

PR #165, opened 2015-09-01
Extra details when a test fails - assertStatusCode

jmclean:
When debugging my controllers I often need more information than isSuccessful currently provides. This PR adds a similar method called assertStatusCode which adds extra information. I don't think I can add this functionality to isSuccessful without changing its signature.

* It listens to kernel.exception rather than trying to read the \<title\> element. The title is fine for HTML responses but one of my projects is a JSON API and it displays errors in JSON format. kernel.exception works for both cases.
* It listens to validation errors, so it can display more information when a form submission fails.

With this assertion, a typical form test might contain:

    $client->submit($form);
    $this->assertStatusCode(302, $client);
    $client->followRedirect();

If it fails, it will fail with an error like:

    Unexpected validation errors:
    + data.email: This value should not be blank.
    + data.username: This value should not be blank.
    
    Failed asserting that 200 matches expected 302.

I've also written a method assertValidationErrors, which just takes an array like `array('data.username', 'data.email')` and outputs a "diff" if the error keys don't match.

Unfortunately this adds a dependency on the symfony validator component. This is also tricky because the old validator component was deprecated in 2.5 and I'm not sure how to support both the old and new versions simultaneously (but if anyone can give me some pointers I can update this PR).

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,20 +66,40 @@ Installation
 Basic usage
 -----------
 
-Use `static::makeClient` to create a Client object. Client is a Symfony class that can simulate HTTP requests to your controllers and then inspect the results. It is covered by the [functional tests](http://symfony.com/doc/current/book/testing.html#functional-tests) section of the Symfony documentation.
+Use `static::makeClient` to create a Client object. Client is a Symfony class
+that can simulate HTTP requests to your controllers and then inspect the
+results. It is covered by the [functional tests](http://symfony.com/doc/current/book/testing.html#functional-tests)
+section of the Symfony documentation.
 
-After making a request, use `assertStatusCode` to verify the HTTP status code. If it fails it will display the last exception message or validation errors encountered by the Client object.
+After making a request, use `assertStatusCode` to verify the HTTP status code.
+If it fails it will display the last exception message or validation errors
+encountered by the Client object.
+
+If you are expecting validation errors, test them with `assertValidationErrors`.
 
 ```php
 use Liip\FunctionalTestBundle\Test\WebTestCase;
 
 class MyControllerTest extends WebTestCase
 {
-    public function testIndex()
+    public function testContact()
     {
         $client = static::makeClient();
-        $crawler = $client->request('GET', '/');
+        $crawler = $client->request('GET', '/contact');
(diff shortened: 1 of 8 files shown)


Test/ValidationErrorsConstraint.php
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    /**
+     * Returns a string representation of the object.
+     *
+     * @return string
+     */
+    public function toString()
+    {
+        return 'validation errors match';
```
lsmith77: what is the purpose of this method?

lsmith77:
can you also add something into the readme?

Test/ValidationErrorsConstraint.php
jmclean: PHPUnit requires it due to PHPUnit_Framework_Constraint extending the PHPUnit_Framework_SelfDescribing interface. I haven't found anywhere that it's called, however.

jmclean:
I'm not great at writing documentation but I gave it a go.

Test/WebTestCase.php
```diff
@@ -691,4 +691,50 @@ public function loginAs(UserInterface $user, $firewallName)
...
+    }
+
+    /**
+     * Assert that the last validation errors within $container match the
+     * expected keys.
+     *
+     * @param array $expected A flat array of field names
+     * @param ContainerInterface $container
+     */
+    public static function assertValidationErrors(array $expected, ContainerInterface $container)
```
lsmith77: why static?

README.md
```diff
@@ -63,6 +63,27 @@ Installation
             storage_id: session.storage.filesystem
     ```
 
+Basic usage
+-----------
```
lsmith77: ``assertValidationErrors()`` is missing here ..

jmclean:
Fair points. Updated. :)

lsmith77:
needs a rebase

jmclean:
Rebased, and code style fixed for the style robot. Want me to combine all these commits?

lsmith77:
nah .. its good.. thanks!

PR #172, opened 2015-10-18
Improve Alice integration

pstephenson02:
This took me quite some time to find: https://github.com/liip/LiipFunctionalTestBundle/issues/143

This will allow users to return their fixtures when calling loadFixtureFiles

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,6 +177,20 @@ Tips for Fixture Loading Tests
     }
     ```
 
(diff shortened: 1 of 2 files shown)


GnafGnaf:
Thanks!

elkangaroo:
Thank you @pstephenson02 this is exactly what we were looking for =)

Issue #180, opened 2015-11-12
Command input is a binary string


aledeg:
I am trying to test the output of one of my command but I can't because the string is in binary format.

Here is a sample of my code:
```php
namespace Acme\Bundle\Tests\Command;

use Liip\FunctionalTestBundle\Test\WebTestCase;

class MyCommandTest extends WebTestCase {

    public function testExecute() {
        $display = $this->runCommand('acme:mycommand');

        $this->assertRegExp('/needle/', $display);
    }
}
```
Unfortunately, the assertion is always false because ```$display``` is a binary string.

I found out that it's caused by the format of the output. By changing the code in the ```WebTestCase``` class I can have the behavior I need.
Here is the original method:
```php
    protected function runCommand($name, array $params = array(), $reuseKernel = false)
    {
        array_unshift($params, $name);

        if (!$reuseKernel) {
            if (null !== static::$kernel) {
                static::$kernel->shutdown();
            }

            $kernel = static::$kernel = $this->createKernel(array('environment' => $this->environment));
            $kernel->boot();
        } else {
            $kernel = $this->getContainer()->get('kernel');
        }

        $application = new Application($kernel);
        $application->setAutoExit(false);

        $input = new ArrayInput($params);
        $input->setInteractive(false);

        $fp = fopen('php://temp/maxmemory:'.$this->maxMemory, 'r+');
        $output = new StreamOutput($fp);

        $application->run($input, $output);

        rewind($fp);

        return stream_get_contents($fp);
    }
```

Here is the modified method:
```php
    protected function runCommand($name, array $params = array(), $reuseKernel = false)
    {
        array_unshift($params, $name);

        if (!$reuseKernel) {
            if (null !== static::$kernel) {
                static::$kernel->shutdown();
            }

            $kernel = static::$kernel = $this->createKernel(array('environment' => $this->environment));
            $kernel->boot();
        } else {
            $kernel = $this->getContainer()->get('kernel');
        }

        $application = new Application($kernel);
        $application->setAutoExit(false);

        $input = new ArrayInput($params);
        $input->setInteractive(false);

        $fp = fopen('php://temp/maxmemory:'.$this->maxMemory, 'r+');
        $output = new StreamOutput($fp, StreamOutput::VERBOSITY_NORMAL, false);

        $application->run($input, $output);

        rewind($fp);

        return stream_get_contents($fp);
    }
```

[...]

PR #182, opened 2015-11-16
Add command output configuration

aledeg:
Before, the StreamOutput object was instanciated with default options.
That will cause some problems when testing the output on terminal
supporting decorated outputs.
Now, it is possible to configure the output to remove the decoration and
choosing the verbosity level.

See #180


Test/WebTestCase.php
```diff
@@ -119,7 +119,19 @@ protected function runCommand($name, array $params = array(), $reuseKernel = fal
         $input->setInteractive(false);
 
         $fp = fopen('php://temp/maxmemory:'.$this->maxMemory, 'r+');
-        $output = new StreamOutput($fp);
+
+        $commandVerbosity = StreamOutput::VERBOSITY_NORMAL;
+        if ($this->getContainer()->hasParameter('liip_functional_test.command_verbosity')) {
```
lsmith77: should we not make these properties instead which are set with the configured defaults? this way we can override them on a per test basis

aledeg: That's a good idea. But we could even have both.
Could you guide me on how to set new properties in that case? I am not sure were to inject those properties.
Thanks

aledeg: I figured how to do that :)

lsmith77:
thx

aledeg:
My pleasure

PR #187, opened 2015-12-03
updated symfony/validator version to 3.0

sc0rp10:


lsmith77:
ok .. merged this .. but I still expect issues to occur due to the deprecated interfaces

Issue #177, opened 2015-10-28
SF 2.5+ compatibility


Elexy:
i just upgraded to symfony 2.7@stable and now see this:
```
Deprecated: The Liip\FunctionalTestBundle\Validator\DataCollectingValidator class implements Symfony\Component\Validator\MetadataFactoryInterface that is deprecated since version 2.5, to be removed in 3.0. Use {@link Mapping\Factory\MetadataFactoryInterface} instead. in /Users/*****/dev/cb_api/vendor/symfony/symfony/src/Symfony/Component/Debug/DebugClassLoader.php on line 208
```

lsmith77:
I would be happy to merge a PR to fix this .. however for now we have to keep 2.3 compatibility. so it might require adding a small BC hack.

alexislefebvre:
Symfony 2.7 has Long Term Support like Symfony 2.3 have. Can't you force users to upgrade to Symfony 2.7 by requiring Symfony ≥ 2.7 ? The upgrade from 2.3 to 2.7 can be done easily, it was developed while keeping ascending compatbility.

Otherwise, can't you create a new branch specific to Symfony ≥ 2.7? It would be available in Composer in parallel to the current 2.3-compatible master branch.

lsmith77:
since its hard enough to find the time to maintain this Bundle, I would like to wait with dropping 2.3 support until when 2.3 stops getting bug fix releases, which is May 2016

alexislefebvre:
@lsmith77 I understand, thanks for your answer.

lakhman:
+1 - this is annoying, updating the validator version like on this fork didn't seem to help, any ideas how I can fix it on my own fork?

https://github.com/hhamon/LiipFunctionalTestBundle/commit/c6f3c35493c95d8c5d890795b60287e04f0fe48b

lsmith77:
I think the issue is now solved.

lakhman:
I've updated to `dev-master` and `dev-master#b5e90734c54d5b28bedd5830439eda3096b9e878`. I'm still seeing the deprecated error, am I doing something wrong?

`composer.json`:
```
   "liip/functional-test-bundle": "dev-master#b5e90734c54d5b28bedd5830439eda3096b9e878",
```

[...]

PR #189, opened 2015-12-05
using assertValidationErrors() requires Symfony 2.5+

lsmith77:
related to https://github.com/liip/LiipFunctionalTestBundle/issues/177

### DIFF
--- a/DependencyInjection/LiipFunctionalTestExtension.php
+++ b/DependencyInjection/LiipFunctionalTestExtension.php
@@ -30,6 +30,9 @@ class LiipFunctionalTestExtension extends Extension
 
         $loader = new XmlFileLoader($container, new FileLocator(__DIR__.'/../Resources/config'));
         $loader->load('functional_test.xml');
+        if (interface_exists('Symfony\Component\Validator\Validator\ValidatorInterface')) {
+            $loader->load('validator.xml');
+        }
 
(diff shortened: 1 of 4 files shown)


lsmith77:
@jmclean can you have a look at this?

jmclean:
Looks good to me. Needs another note in the readme that assertStatusCode also requires version 2.5 (in addition to assertValidationErrors). Then the `symfony/validator` constraint can be removed from composer.json and it's good to go!

alexislefebvre:
I think `if (interface_exists('Symfony\Component\Validator\Validator\ValidatorInterface')) {` is used to test that Symfony 2.5+ is used. But this interface exists with the last version of Symfony 2.3 and I have an error when I launch tests, it seems that `$loader->load('validator.xml');` is called with Symfony 2.3 too. The error in the following test seems to be caused by this code: https://travis-ci.org/alexislefebvre/LiipFunctionalTestBundle/jobs/97578616

jmclean:
I'm pretty sure that error is coming from the constraint on symfony/validator in this bundle's composer.json. It's resulted in you getting version 2.6 of the validator but version 2.3 of the symfony core. @lsmith77 the symfony/validator line can just be removed from composer.json now. It's indirectly required by symfony/framework-bundle and the version isn't a strict requirement any more.

PR #190, opened 2015-12-07
Fix #178 - Move to the new symfony validator signature

### LINKED ISSUE #178: Conflicts with sonata admin bundle 2.3
So, I installed 1.3-dev version of LiipFunctionalTestBundle and after reloading project got this exception:

`
FileLoaderLoadException: Catchable Fatal Error: Argument 1 passed to Sonata\AdminBundle\Admin\Admin::setValidator() must implement interface Symfony\Component\Validator\ValidatorInterface, instance of Liip\FunctionalTestBundle\Validator\DataCollectingValidator given, called in /var/www/app.gyman/app/cache/dev/appDevDebugProjectContainer.php on line 713 and defined in /var/www/app.gyman/app/config/. (which is being imported from "/var/www/app.gyman/app/config/routing.yml").
`

resilva:
A try to fix the #178 issue, seems working for my use

### DIFF
--- a/Validator/DataCollectingValidator.php
+++ b/Validator/DataCollectingValidator.php
@@ -16,7 +16,7 @@ use Symfony\Component\HttpKernel\KernelEvents;
 use Symfony\Component\Validator\ConstraintViolationList;
 use Symfony\Component\Validator\ConstraintViolationListInterface;
 use Symfony\Component\Validator\Context\ExecutionContextInterface;
-use Symfony\Component\Validator\Validator\ValidatorInterface;
+use Symfony\Component\Validator\ValidatorInterface;
 
 class DataCollectingValidator implements ValidatorInterface, EventSubscriberInterface
 {
@@ -56,7 +56,12 @@ class DataCollectingValidator implements ValidatorInterface, EventSubscriberInte
         return $this->wrappedValidator->hasMetadataFor($value);
     }
 
-    public function validate($value, $constraints = null, $groups = null)
+    public function validate($object, $groups = null, $traverse = false, $deep = false)
+    {
+        return $this->wrappedValidator->validate($object, $groups, $traverse = false, $deep = false);
+    }
+
+    public function validateValue($value, $constraints = null, $groups = null)
     {
         return $this->lastErrors = $this->wrappedValidator->validate($value, $constraints, $groups);
     }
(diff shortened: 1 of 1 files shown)


luispabon:
:+1:

Validator/DataCollectingValidator.php
```diff
@@ -16,7 +16,7 @@
 use Symfony\Component\Validator\ConstraintViolationList;
 use Symfony\Component\Validator\ConstraintViolationListInterface;
 use Symfony\Component\Validator\Context\ExecutionContextInterface;
-use Symfony\Component\Validator\Validator\ValidatorInterface;
+use Symfony\Component\Validator\ValidatorInterface;
```
lsmith77: isn't this the old interface?

https://github.com/symfony/validator/blob/master/Validator/ValidatorInterface.php

luispabon: Indeed. Need to bump min symfony version to 2.6 then use the new interface.

resilva:
You're right @lsmith77 it's the old signature, the commit message is not good. Indeed it's still solve the problem of compatibility with sonata, the best would be as suggested by @luispabon to upgrade Symfony and set all the things clear.

It's still up to you to decide how you want the problem to be solved, I would be happy to help to improve this pullrequest.

luispabon:
You can also boilerplate-code your way around being compatible with both interfaces, I personally wouldn't and just warn on the docs which version of LFTB to install for those projects which won't update to symfony 2.6 for some reason.

lsmith77:
I am fine with requiring higher versions of Symfony for this feature

luispabon:
Any joy with this? Maybe I can complete this PR?

resilva:
Sorry I didn't take the time to correct my pr, I think I will rollback what I've done and fix the minimal required version of Symfony as we discussed earlier.

I will fix this next week if anyone want to fix it before go for it ;)

luispabon:
:muscle:

resilva:
Sorry too much work this week to handle this, I will try on January since I'm on holiday :(

lsmith77:
see master

elkangaroo:
Can you reference the commit, that fixes this? I don't seem to find it (:blind:).

alexislefebvre:
@elkangaroo I think it's change: https://github.com/liip/LiipFunctionalTestBundle/commit/d3c9977288031ee01849515369a5c91be1b1474c#diff-ab2cc27d1cdfd485cf44b42243abdc9a

PR #195, opened 2015-12-18
Fixes #194: use PHP5.3-compatible array syntax

### LINKED ISSUE #194: Keep PHP 5.3 compatibility in Test/WebTestCase.php
The changes from #193 use short array syntax from PHP 5.4, This code fails with PHP 5.3: https://travis-ci.org/alexislefebvre/LiipFunctionalTestBundle/jobs/97735960#L285

It uses the short array syntax which appeared with PHP 5.4: http://php.net/manual/en/migration54.new-features.php

Since LiipFunctionalTestBundle is compatible with Symfony 2.3 which is compatible with PHP 5.3, the *old* array syntax should be used.

This is similar to this previous issue: #149.

alexislefebvre:
instead of short array syntax

### DIFF
--- a/Test/WebTestCase.php
+++ b/Test/WebTestCase.php
@@ -585,10 +585,10 @@ abstract class WebTestCase extends BaseWebTestCase
                 $authentication = $this->getContainer()->getParameter('liip_functional_test.authentication');
             }
 
-            $params = array_merge($params, [
+            $params = array_merge($params, array(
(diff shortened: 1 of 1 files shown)


lsmith77:
hmm .. isn't this redundant to https://github.com/liip/LiipFunctionalTestBundle/commit/0981fc0a18678b50a53410496239b602dc5a355b .. strange that github doesn't see this as a conflict.

PR #199, opened 2015-12-28
Ensure minimum requirement when PHP 5.4.*

ericpoe:
Prior to Doctrine-common 2.6, the minimum version of PHP required is [PHP 5.3.2](https://github.com/doctrine/common/blob/2.5/composer.json). With Doctrine-common 2.6, the version of PHP required is PHP [5.5](https://github.com/doctrine/common/blob/2.6/composer.json). So, for those of us with PHP 5.4.*, when we run `composer update --dry-run`, we see that LiipFunctionalTestBundle 1.3.3 needs to be updated; yet, when updated, our tests that rely upon LiipFunctionalTestBundle fails.

Doctrine Common 2.6 adds the requirement for PHP >= 5.5.x. For projects
that still have PHP 5.4.x, this causes many problems in testing when
upgrading LiipFunctionalTestBundle from 1.2.2 to 1.3.3.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
         "symfony/framework-bundle": "~2.3|~3.0",
         "symfony/browser-kit": "~2.3|~3.0",
         "symfony/validator": "~2.5|~3.0",
-        "doctrine/common": "~2.0"
+        "doctrine/common": "^2.5.3"
     },
     "suggest": {
         "doctrine/doctrine-fixtures-bundle": "Required when using the fixture loading functionality",


lsmith77:
I do not quite understand the issue. Is something broken in this Bundle or are you just seeing stuff break because you are upgrading Doctrine dependencies?

ericpoe:
Stuff is breaking. New versions of LiipFunctionalTestBundle utilize Doctrine-common 2.6 and don't explicitly require a minor release version (2.5 vs 2.6). Two options I see to ensure that tests that use this bundle do not break on upgrade are to either explicitly lower the required doctrine-common version or raise the required PHP version from 5.3.9 to 5.5.

lsmith77:
uhm this sounds like a regression in doctrine common .. can you paste or explain in more detail the issues that the newer doctrine common version creates? and then I guess you should instead open a ticket on their repo

/cc @Ocramius

Ocramius:
> this causes many problems in testing when upgrading LiipFunctionalTestBundle from 1.2.2 to 1.3.3.

Sorry, but I think this is obligatory :-P

tmp_21825-bart-simpson-generator php-863659251

lsmith77:
your problem is that you are using different PHP versions in development and production. This is something you need to solve on your end. There are ways to force composer always to target PHP 5.4 when doing updates. But as @Ocramius said .. its not a good idea to have your dev and prod env not in sync.

ericpoe:
@lsmith77 & @Ocramius, my versions of PHP in prod and dev are the same: PHP 5.4. Where are you seeing that they are different?

Since this PR is closed, I'll see myself out. Cheers.

Ocramius:
>  when we run composer update --dry-run, we see that LiipFunctionalTestBundle 1.3.3 needs to be updated; yet, when updated, our tests that rely upon LiipFunctionalTestBundle fails.

If you run a `composer update` from any environment with 5.5, then common 2.6 will be installed. Upon deployment, you will see the crashes.

ericpoe:
Vagrant box is running 5.4. Prod is running 5.4. Local machine is 5.6, so you got me. However, updates and testing are run from Vagrant box.

It's probably something I'm doing. If this comes up again, I'll create a ticket. Cheers.

Ocramius:
@ericpoe thanks!
And sorry if I sounded like an ass, I just had a few similar issues reported the same day, and should have answered in a more polite way initially :-\

lsmith77:
@ericpoe you can try something like the following in your composer.json (that I mentioned above):

```

    "config": {
        "platform": {
            "php": "5.4.0"
        }
    }
```

this way composer will always behave as if php 5.4.0 was your installed php version

ericpoe:
Thanks, @lsmith77. I had added that config to my `composer.json` at some point during my upgrade of Symfony & libraries -- perhaps as a trouble-shooting step for this problem. :)

No worries @Ocramius. We all have our off days. Just don't make it a habit of biting off the head of someone who is trying to help you.

Happy New Year!

PR #203, opened 2016-01-07
cleaned up some of the fixture related code, minor BC break

### LINKED ISSUE #135: getEntityManager is deprecated since Symfony 2.1
"Deprecated: getEntityManager is deprecated since Symfony 2.1. Use getManager"

see:

https://github.com/liip/LiipFunctionalTestBundle/blob/master/Test/WebTestCase.php#L242

lsmith77:
fixes #135

due to the minor BC break I guess we should at least bump the minor version


Test/WebTestCase.php
```diff
@@ -212,22 +213,6 @@ protected function getContainer()
     }
 
     /**
-     * @param string $omName
-     * @param string $registryName
-     *
-     * @return ObjectManager
-     */
-    protected function getObjectManager($omName = null, $registryName = 'doctrine')
```
lsmith77: not sure why this method was ``protected`` to begin with .. and I doubt anyone has used this method as an extension point

PR #198, opened 2015-12-21
Fixes #197: Added tests and Continuous Integration services:

alexislefebvre:
Added test environment:
 - AppKernel with configuration and route (see http://www.whitewashing.de/2012/02/25/symfony2_controller_testing.html for reference)
 - Dependencies in composer.json: Symfony and PHPUnit

Added DataFixtures and tests with fixtures
Added User entity (I had to use the YAML format for Doctrine, annotations didn't worked)
Web tests: added a form with validation errors assertions, added status code assertions
Added:
 - Simple command and a test for it
 - Twig view
 - Simple test file with fake authentification

Travis CI: Symfony 2.3 is in allowed failures
Scrutinizer: Composer will install Symfony 2.8.*

Added badges in README.md: Travis CI, Scrutinizer and SensioLabsInsight

You will have to update the URLs in README.md once CI services will be configured for this repository.

PR #204, opened 2016-01-08
Tests: improved command and form tests

alexislefebvre:
Fixed call to assertValidationErrors()
Added NotBlank constraint to the form
Added different commands calls (improve code coverage)


alexislefebvre:
That's strange, squashing and forced push worked, contrary to my previous PR #198.

PR #211, opened 2016-01-11
Tests: Added HTML5 validation

alexislefebvre:
This will improve code coverage.

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -50,6 +50,12 @@ return $v; })
                         ->scalarNode('password')->end()
                     ->end()
                 ->end()
+                ->arrayNode('html5validation')
+                    ->children()
+                        ->scalarNode('url')->end()
+                        ->arrayNode('ignores_extract')->end()
+                    ->end()
+                ->end()
             ->end()
         ;
 
--- a/Resources/config/functional_test.xml
+++ b/Resources/config/functional_test.xml
@@ -11,6 +11,8 @@
         <parameter key="liip_functional_test.html5validation.url">http://validator.nu/</parameter>
 
         <parameter key="liip_functional_test.html5validation.ignores" type="collection" />
+        
+        <parameter key="liip_functional_test.html5validation.ignores_extract" type="collection" />
 
         <parameter key="liip_functional_test.authentication" type="collection">
             <parameter key="username"></parameter>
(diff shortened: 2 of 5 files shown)


DependencyInjection/Configuration.php
```diff
@@ -50,6 +50,12 @@ public function getConfigTreeBuilder()
                         ->scalarNode('password')->end()
                     ->end()
                 ->end()
+                ->arrayNode('html5validation')
+                    ->children()
+                        ->scalarNode('url')->end()
+                        ->arrayNode('ignores_extract')->end()
```
lsmith77: can you add something to the docs for this?

alexislefebvre: It was needed because of this line: https://github.com/liip/LiipFunctionalTestBundle/pull/211/files#diff-da25f5e5e9255b14c3d3d96edd20bb24R154

It only define nodes already used in the tests: https://github.com/alexislefebvre/LiipFunctionalTestBundle/blob/html5-validator/Test/Html5WebTestCase.php#L69

lsmith77: ah ok.

alexislefebvre: Anyway you're right, the `html5validation` node is not documented in the project's README, I'll try to improve this and submit a PR.

alexislefebvre: Should the `liip_functional_test.html5validation.ignores` and `liip_functional_test.html5validation.ignores_extract` parameter be kept? It looks like it's used to ignore validation errors. These lines haven't been updated since 2011. Should I open an issue to talk about removing unused features?

alexislefebvre: Related: #213.

lsmith77:
hmm .. strange .. build is failing now:
https://travis-ci.org/liip/LiipFunctionalTestBundle/jobs/102196658

alexislefebvre:
@lsmith77 it must be due to the fact that the service is unavailable at this moment:

    $ wget https://validator.nu/ -O-
    --2016-01-13 22:09:19--  https://validator.nu/
    Resolving validator.nu (validator.nu)... 46.226.110.49
    Connecting to validator.nu (validator.nu)|46.226.110.49|:443... connected.
    HTTP request sent, awaiting response... 502 Bad Gateway
    2016-01-13 22:09:19 ERROR 502: Bad Gateway.

I'll add a check to skip the test for this case.

alexislefebvre:
The `validateHtml5()` function return the result of the request or `false`, but the test `if (false === $res->messages) {` fail because the key `messages` is invalid if `false` is returned.

PR #213, opened 2016-01-12
Updated validator's URL added in #211: http://validator.nu/ now uses HTTPS

alexislefebvre:
Added test of code snippet
Revert some unnecessary changes from #211 and cc2c9efc64d09cace82e701deb44bb071ab6b6b6

PR #215, opened 2016-01-13
Tests: Skip validator tests if the response is invalid

alexislefebvre:
ests: Skip tests from Html5WebTestCase if the validator service didn't return JSON content, see #211 and #213


alexislefebvre:
We can also check that `$res->messages` is defined:

> The root object is a JSON object. It has one mandatory key, `"messages"`, and three optional keys, `"url"`, `"source"` and `"parseTree"`.

Source: https://github.com/validator/validator/wiki/Output:-JSON#root-object

PR #223, opened 2016-01-14
Tests: HTML validator: use fake data instead of calling the web service

alexislefebvre:
Remove skippable tests in order to restore code coverage
The Bundle won't depend on the validator service during tests


alexislefebvre:
It should prevent this type of artifacts:

test coverage changes - inspection of merge pull request 215 from alexislefebvre-html5- - liip-liipfunctionaltestbundle - scrutinizer_

Issue #224, opened 2016-01-14
Add `tearDown()` method to `WebTestCaseTest` (Enhancement)


Aerendir:
Set `$this->client` to `null`.

alexislefebvre:
Mhh, you changed the test which test the Bundle itself, not the actual code used when the Bundle is used as a dependency, see the [`Test`](https://github.com/liip/LiipFunctionalTestBundle/tree/master/Testà directory.

Aerendir:
Yes, I changed the test...

alexislefebvre:
I'm sorry, I thought it was an error. I tried your code and it didn't changed the memory usage reported by PHPUnit. Do you see a difference?

Aerendir:
@alexislefebvre , yes, it didn't changed the memory usage... Maybe... But this is a best practice, and while currently it didn't impact much more, in the future it'll may.

Is something related to the approach: if we "dirty" the tests, we have to clean. Viceversa, in the long run we will see a lot of garbage not collected.

The besta example is `WebTestCase` that consumes a lot of memory while, using the `tearDown()` method, we could free up it and speed up the tests.

Issue #218, opened 2016-01-14
Use private properties to store sensible and big variables to improve memory usage


Aerendir:
I'm refactoring some of my test class to use less memory as I had some problems with large memory usage on my CI server.

Going deeper in the code to find issues I came up with a `trait` that can teardown automatically the tests and set all properties to `null` to free up memory space.

The `trait` also provides some useful methods to better manage internal variables and resources, storing them in some properties that are then set to `null`.

This bundle seems to use a lot of memory as it creates a lot of very big variables.

It may be sufficient to use properties to store objects like kernel, clients, and similar thing, and, adding a simple `tearDown()` method, set those to `null` to free up memory space and improve tests performance.

Maybe this helpful to you?

lsmith77:
yeah .. I think the biggest memory hog is the container etc .. which I think leaks memory .. sure send a PR!

Aerendir:
Perfect! I'll work on this :)

tackerm:
Yesss, +1 on improving memory usage!

You may want to take a look at this: http://kriswallsmith.net/post/18029585104/faster-phpunit
We do something similar in our tests, and it helps a lot!

It also helps to force a garbage collection cycle in the tearDown() method (however, if you are using the phpunit-bridge note that the bridge currently disables garbage collection).

alexislefebvre:
IIRC, this Bundle inherits from Symfony test class, is the `tearDown()` method from Symfony implements this optimisation?

tackerm:
It just shuts down the kernel in `KernelTestCase::tearDown()`, but does not clean up anything else.

alexislefebvre:
Maybe it would be interesting to submit the same patch to Symfony?

tackerm:
Yes and no.
In terms of clean coding, every class should take care of cleaning up after itself. So every `*Test`-class should have a `tearDown()` method that removes/unsets all the properties created by that class.

The Symfony test classes do not set any properties (except for the kernel, which is also being shut down). The `Liip\FunctionalTestBundle\Test\WebTestCase` class however should take care of properly unsetting the `$environment, $containers, $kernelDir, $maxMemory, $verbosityLevel, $decorated` properties.

The patch mentioned above is basically just a shortcut for the lazy developer - I'm not sure if this should be provided/encouraged by the framework :)

Aerendir:
[...]

PR #226, opened 2016-01-15
#224. Add `tearDown()` method to `WebTestCaseTest`.

Aerendir:


alexislefebvre:
Is it related to #218?

Aerendir:
@tackerm , read this about the difference between the two: http://stackoverflow.com/questions/584960/whats-better-at-freeing-memory-with-php-unset-or-var-null

It seems that set the variable to `null` is a it faster than `unset` it.

tackerm:
This may now be an academic excursion, but:

Setting a variable to `null` is 
* "faster" in terms of freeing up *some* memory *immediately*
* "slower" in terms of overall code execution (because of the additional cycles)

`unset`ing a variable is
* "better" in terms of freeing up *all* memory *eventually* (at the next gc cycle -> garbage collection must be enabled)
* "faster" in terms of overall code execution

In my own projects, memory consumption is **always** the limiting factor - I think the code execution time is neglectible (especially for tests), but the *some memory* vs. *all memory* issue can make quite a difference :wink: 

Anyways, any kind of cleanup is better than none, so :+1: :smile:

Aerendir:
@tackerm , yes, maybe a simple academic excursion, but I'm interested in it as We should choose a way (and I have to choose one for my projects, too! :) ).

What do you prefer? Maybe we could use both of them?

tackerm:
Well, this specific PR only affects the test which tests the bundle itself, so it doesn't really matter. But if you plan to add a `tearDown()` to `WebTestCase`, then I'd strongly recommend to use `unset()`.

Aerendir:
:+1: Then I'll use `unset()`... I were on the same direction...

lsmith77:
also the PR needs a rebase.

Aerendir:
@lsmith77 , the PR is now ready!

Aerendir:
It seems Scrutinizer failed in something... Is this a fault of mine? I think it isn't, but I may be wrong...

lsmith77:
retriggered. lets see

Issue #229, opened 2016-01-15
`WebTestCase::retrieveVerbosityLevel()` always returns StreamOutput::VERBOSITY_NORMAL (Bug)


Aerendir:
I'm trying to use `WebTestCase::runCommand()` but it ever returns an empty output while I'm expecting something.

In fact, to get the output I need, I have to set the `verbosity_level` to `debug`.

But the method always returns `StreamOutput::VERBOSITY_NORMAL`.

Going deeper in the issue, I found the bug is in the method `WebTestCase::getVerbosityLevel()` that works this way:

```php
private function retrieveVerbosityLevel()
    {
        // Returns the local verbosity level
        if ($this->verbosityLevel) {
            $verbosity = 'StreamOutput::VERBOSITY_'.strtoupper($this->verbosityLevel);
            if (defined($verbosity)) {
                return $verbosity;
            }
        }

        // Returns the global verbosity level
        if ($this->getContainer()->hasParameter('liip_functional_test.command_verbosity')) {
            $verbosity = 'StreamOutput::VERBOSITY_'.strtoupper($this->getContainer()->getParameter('liip_functional_test.command_verbosity'));
            if (defined($verbosity)) {
                return $verbosity;
            }
        }

        // Returns the default verbosity level
        return StreamOutput::VERBOSITY_NORMAL;
    }
```

The problem is in the checks `if (defined($verbosity))` that seems ever return `false`.

In fact, trying to print some breakthrough, the method ever fallbacks to the default `verbosity_level` while I'm configuring it as `debug` (I've tried to configure it both as local with `$this->verbosityLevel = 'debug'` and as global, by setting in the `config_test.yml` the key `liip_functional_test.command_verbosity`).

This is the modified method I've used to discover the problem:

```php
private function retrieveVerbosityLevel()
    {
        // Returns the local verbosity level
        if ($this->verbosityLevel) {
            printf("Local verbosity is set to %s", $this->verbosityLevel);
            $verbosity = 'StreamOutput::VERBOSITY_'.strtoupper($this->verbosityLevel);
            if (defined($verbosity)) {
                return $verbosity;
            } else {
                printf(" but the constant doesn't exist.\n", $this->verbosityLevel);
            }
        }

        // Returns the global verbosity level
        if ($this->getContainer()->hasParameter('liip_functional_test.command_verbosity')) {
[...]

PR #230, opened 2016-01-15
#229. Use `constant()` to check if the verbosity `CONSTANT` exists

Aerendir:


lsmith77:
i guess this fixes https://github.com/liip/LiipFunctionalTestBundle/issues/229
would be good to add a test case ..

Aerendir:
I'll write one now...

Aerendir:
I've wrote the tests, but as `retrieveDecorated` 1) isn't accessible, 2) doesn't use a property and `runCommand()` doesn't return an `OutputInterface` object, it is practically impossible to check it's value. :(

But I've added a lot of tests to check a lot of conditions that should be respected.

Aerendir:
I've added a check for the Symfony version and fix the fails of the tests.
But remains failures on PHP 5.3 and sincerely I have no idea about the cause. @lsmith77  and @alexislefebvre any ideas?

Aerendir:
I need help: on PHP 5.3 tests continue to fail. It seems that the setting of verbosity is completely ignored and I really can't understand why. I've also added a setter for verbosity, refactored the `retrieveVerbosityLevel()` method (I called it `getVerbosityLevel()`), done a version check about symfony version (that solved some previous fails on some server configurations) but with PHP 5.3 and Symfony 2.3.1 the value is ever set to `VERBOSITY_NORMAL` and I really don't understand why.

@lsmith77 and @alexislefebvre , can you check? Maybe I'm missing something... Thank you!

alexislefebvre:
I don't know sorry. Have you check that this verbosity level exists in Symfony 2.3.1? Maybe it has been added later.

Aerendir:
Solved! After hours of deep inspections and `var_dumps`, all now works well.

Symfony 2.3.1 has a wrong behavior in handling verbosity levels, so I adjusted tests consequently and added some checks on symfony version to make all work on every environment.

Test/WebTestCase.php
```diff
@@ -118,11 +121,15 @@ protected function runCommand($name, array $params = array(), $reuseKernel = fal
         $application = new Application($kernel);
         $application->setAutoExit(false);
 
+        if (\Symfony\Component\HttpKernel\Kernel::VERSION_ID === '20301') {
```
lsmith77: can you add a use statement for the ``Kernel`` class?

Test/WebTestCase.php
```diff
@@ -134,30 +141,76 @@ protected function runCommand($name, array $params = array(), $reuseKernel = fal
...
-        if ($this->verbosityLevel) {
-            $verbosity = 'StreamOutput::VERBOSITY_'.strtoupper($this->verbosityLevel);
-            if (defined($verbosity)) {
-                return $verbosity;
-            }
+        // If `null`, is not yet set
+        if (null === $this->verbosityLevel) {
+            // Set the global verbosity level that is set as NORMAL by the TreeBuilder in Configuration
+            $level = strtoupper($this->getContainer()->getParameter('liip_functional_test.command_verbosity'));
+            $verbosity = '\Symfony\Component\Console\Output\StreamOutput::VERBOSITY_'.$level;
```
lsmith77: can you add a use statement for ``StreamOutput`` ?

alexislefebvre:
@Aerendir I should have think about this before but you may have to chosen to skip the tests with Symfony 2.3, that's already the case with some tests, e.g. https://travis-ci.org/liip/LiipFunctionalTestBundle/jobs/102919215

@lsmith77 will be able to tell us if it's better to write specific code for some versions or instead skip these tests.

Test/WebTestCase.php
Aerendir: Yes, here I can add a use statement.

Test/WebTestCase.php
Aerendir: No, here it isn't possible to use the `use` statement as the `constant()` function cannot "intercept" it. The full namespace is needed to make the code work.

Aerendir:
@alexislefebvre , don't worry: in the end I solved the problem and learned something more :)

Aerendir:
@lsmith77 , the PR now is ready!

PR #233, opened 2016-01-16
Tests: added Configuration tests

alexislefebvre:
Values are different from the default values, this test check that they were loaded.


alexislefebvre:
I hope that the [configuration used during tests](https://github.com/liip/LiipFunctionalTestBundle/blob/7950455834d87fb1722d18731ed8f99f9ec499d5/Tests/App/config.yml#L31) don't have side effects. For example, if the Bundle tests work but the Bundle is broken when integrated in another project.

It would be perfect if the Bundle could also load a `config.yml` file with only the `liip_functional_test: ~` and run some tests which work with the default configuration. It would ensure than the Bundle can be used with the default configuration.

lsmith77:
different config files can likely best be done with different environments and then multiple calls to PHPUnit and then making sure that travis fails if either of the PHPUnit calls fail ..

alexislefebvre:
Yes, it's possible to create different `phpunit.xml` files in order to load different `AppKernel.php` files, from where the [configuration is loaded](https://github.com/liip/LiipFunctionalTestBundle/blob/3e4e8921b86eebf9a016d941c8ec824282564ca7/Tests/App/AppKernel.php#L38). It's not easy though.

I found that SwitMailer tests load different configuration files with a [`loadContainerFromFile()` method](https://github.com/symfony/swiftmailer-bundle/blob/master/Tests/DependencyInjection/SwiftmailerExtensionTest.php#L296). But it looks like the tests only check the loaded configuration, not the actual behaviours of the Bundle.

alexislefebvre:
I added a the `AppConfigKernel.php` file which load a new `config.yml` which use all the configuration nodes defined by this Bundle. I had to use PHPUni's annotation `@runTestsInSeparateProcesses` to load this new Kernel and not interfer with the default parameters.

I have updated `DependencyInjection/Configuration.php` to define default values with `->addDefaultsIfNotSet()` instead of relying on `Resources/config/functional_test.xml`.

This add a lot of files, I'll try to optimize it.

alexislefebvre:
I've finished to work on it. It needs to be reviewed now. :)

Aerendir:
@alexislefebvre , I removed the check on `command_verbosity` and `command_decoration`. Those default values should be set by the `TreeBuilder` as they are ever needed. This will not force the developer to set them before using `runCommand()` as the method has all the values required to work.

alexislefebvre:
@Aerendir OK, I'll adapt the tests once your PR will be merged.

Aerendir:
:+1:

lsmith77:
https://github.com/liip/LiipFunctionalTestBundle/pull/230 is merged

alexislefebvre:
Thanks for your messages. This PR is now synced with the master branch.

Issue #234, opened 2016-01-21
Move directories used only in tests in Tests/App/ directory


alexislefebvre:
The `Controller`, `Command`, `DataFixtures` and `Entity` directories added for testing this Bundle may be moved in the `Tests` directory in order to keep the directory root cleaner.

alexislefebvre:
Moving `Entity` requires more work: http://stackoverflow.com/questions/18503093/how-do-i-change-symfony-2-doctrine-mapper-to-use-my-custom-directory-instead-of/18515797#18515797

alexislefebvre:
It's possible to move the `Command` directory but it's not available with Symfony 2.3, according to the [documentation][1] it's possible only with Symfony 2.4+

[1]: http://symfony.com/doc/2.7/cookbook/console/commands_as_services.html

Aerendir:
The occasion may be good to create also a subfolder `src` for the bundle... Just a suggestion to keep things better organized...

alexislefebvre:
@Aerendir can you please explain what are pros and cons of the `src/` subfolder?

I don't remember having seen a Bundle with a `src/` subfolder. And the [Symfony  Best Practices for Reusable Bundles ][1] don't use a `src/` subfolder, IMHO we need strong arguments in order to go against these practices.

[1]: http://symfony.com/doc/2.7/cookbook/bundles/best_practices.html#directory-structure

Aerendir:
@alexislefebvre , yes, you are right... As not told! :)

alexislefebvre:
Closed with #235.

PR #235, opened 2016-01-21
Move directories used only in tests in Tests/App/ directory

alexislefebvre:
See #234.


alexislefebvre:
What about a dedicated directory in `Tests` to store everything which define the Bundle used during tests?

PR #236, opened 2016-01-23
Fix bugs with array parameters

alexislefebvre:
Parameters with array nodes are not directly accessible,
the parent node must be declared in PHP before accessing the relevant array key

### DIFF
--- a/Tests/Test/WebTestCaseTest.php
+++ b/Tests/Test/WebTestCaseTest.php
@@ -12,7 +12,6 @@
 namespace Liip\FunctionalTestBundle\Tests\Test;
 
 use Liip\FunctionalTestBundle\Test\WebTestCase;
-use Liip\FunctionalTestBundle\Annotations\QueryCount;
 
 class WebTestCaseTest extends WebTestCase
 {
@@ -324,9 +323,6 @@ class WebTestCaseTest extends WebTestCase
         );
     }
 
-    /**
-     * @QueryCount(100)
-     */
     public function testUserWithFixtures()
     {
         $fixtures = $this->loadFixtures(array(
--- a/DependencyInjection/Compiler/SetTestClientPass.php
+++ b/DependencyInjection/Compiler/SetTestClientPass.php
@@ -10,7 +10,12 @@ class SetTestClientPass implements CompilerPassInterface
 {
     public function process(ContainerBuilder $container)
     {
-        if (null === $container->getParameter('liip_functional_test.query.max_query_count')) {
+        // "query.max_query_count" is an array, it is only accessible
+        // through "query" node and getting the "max_query_count" array
+        // key with PHP.
+        $query = $container->getParameter('liip_functional_test.query');
+
+        if (null === $query['max_query_count']) {
             $container->removeDefinition('liip_functional_test.query.count_client');
 
             return;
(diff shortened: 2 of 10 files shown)


alexislefebvre:
I think this PR requires the smallest amount of work to fetch `liip_functional_test.query.max_query_count`, `liip_functional_test.html5validation.ignores` and `liip_functional_test.html5validation.ignores_extract`.

Tests/Test/WebTestCaseTest.php
```diff
@@ -324,9 +323,6 @@ public function testLoadFixturesFilesPaths()
         );
     }
 
-    /**
-     * @QueryCount(100)
```
lsmith77: do we then still test the annotation some where?

alexislefebvre: Now yet, I tried but I get an error:

> 1) Liip\FunctionalTestBundle\Tests\Test\WebTestCaseConfigTest::testAdminAuthenticationLoginAs
> Doctrine\Common\Annotations\AnnotationException: [Semantical Error] The annotation "@Liip\FunctionalTestBundle\Annotations\QueryCount" in method Liip\FunctionalTestBundle\Tests\Test\WebTestCaseConfigTest::testAdminAuthenticationLoginAs() does not exist, or could not be auto-loaded.

It looks like the annotation reader break some things, I have errors with PHPUnit's annotations too:

> 1) Liip\FunctionalTestBundle\Tests\Test\WebTestCaseConfigTest::testAdminAuthenticationInMemoryRoleAdmin
> Doctrine\Common\Annotations\AnnotationException: [Semantical Error] The annotation "@depends" in method Liip\FunctionalTestBundle\Tests\Test\WebTestCaseConfigTest::testAdminAuthenticationInMemoryRoleAdmin() was never imported. Did you maybe forget to add a "use" statement for this annotation?

alexislefebvre: Here is an example with this branch: https://github.com/alexislefebvre/LiipFunctionalTestBundle/tree/fix-bug-with-array-parameters-annotation-error

Tests fail: https://travis-ci.org/alexislefebvre/LiipFunctionalTestBundle/jobs/104322732

Maybe there is a better way to read annotations from a Bundle? http://stackoverflow.com/questions/29089757/can-i-implement-my-own-symfony2-annotations-easily/29095296#29095296

alexislefebvre: @lsmith77 I don't know anything about services and annotations reader, if nobody can fix this annotation problem, I suggest to remove this annotation feature. Did it ever worked? Has it been broken after a Symfony update?

There's still a way to check the number of queries for a specific request: https://github.com/liip/LiipFunctionalTestBundle/blob/72c8feff5fe6765e33b58397d5cbc98923d3b4b9/Tests/Test/WebTestCaseTest.php#L343

alexislefebvre:
After hours of search, I finally found a solution: http://stackoverflow.com/questions/11013387/doctrine-common-trying-to-create-an-annotation/31519723#31519723 The problem didn't come from the code but from the environment test.

The commit I'm going to push should give a nice increase to the code coverage report from Scrutinizer!

alexislefebvre:
I fixed the bug with the annotations in a already pushed commit: https://github.com/liip/LiipFunctionalTestBundle/commit/d1a2ca1c661f83022868e78603b948dc8c0153ad

This PR only change how the array parameters are read.

alexislefebvre:
I added a message if the profiler is disabled, it's a little bit dirty but I think it's better to show a message than pass the test silently without checking the query count.

PR #237, opened 2016-01-27
Add methods from WebTestCase.php in documentation

alexislefebvre:
isSuccessful()
assertStatusCode()
fetchCrawler()
fetchContent()
getURL()

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,6 +111,98 @@ class MyControllerTest extends WebTestCase
 }
 ```
 
+### Methods
+
+#### Check HTTP status codes
+
+##### isSuccessful()
+
+Check that the request succedded:
+
+```php
+$client = static::makeClient();
+$client->request('GET', '/contact');
+
+// Successful HTTP request
+$this->isSuccessful($client->getResponse());
+```
+
+Add `false` as the second argument in order to check that the request failed:
+
+```php
+$client = static::makeClient();
+$client->request('GET', '/error');
+
+// Request returned an error
+$this->isSuccessful($client->getResponse(), false);
+```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -111,6 +111,98 @@ class MyControllerTest extends WebTestCase
...
+
+// Request returned an error
+$this->isSuccessful($client->getResponse(), false);
+```
+
+In order to test more specific status codes, use `assertStatusCode()`:
+
+##### assertStatusCode()
+
+Check the [HTTP status code](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes) from the request:
```
SofHad: I think that the link to wikipedia is not very useful because ```HTTP status code``` is a basic knowledge for a ```Symfony``` developer :-)

SofHad:
+1

README.md
alexislefebvre: OK, I'll remove it.

alexislefebvre:
In fact, `assertStatusCode()` is already documented in the [current version][1]:

> After making a request, use assertStatusCode to verify the HTTP status code. If it fails it will display the last exception message or validation errors encountered by the Client object.

[1]: https://github.com/liip/LiipFunctionalTestBundle/blob/9ae7294d813f2978bf7e7d758cb1a876de129497/README.md#basic-usage

Is it redundant with the paragraph I added about `assertStatusCode()`? Should I remove this paragraph?

Aerendir:
@alexislefebvre I think when speaking about documentation, nothing is never redundant! :+1:

PR #241, opened 2016-02-03
Simpler way to define array parameters

alexislefebvre:
Define array nodes in configuration as sub-nodes
instead of defining parent node as an array

Eg.
"liip_functional_test.query.max_query_count"
will be a node instead of setting an array
"liip_functional_test.query" with a "max_query_count" key.

Revert some changes from #236

Use idea from FOSUserBundle:
https://github.com/FriendsOfSymfony/FOSUserBundle/blob/d3fc0f8b24ff94a8b5d5888f2264853819ae707b/DependencyInjection/FOSUserExtension.php#L238

### DIFF
--- a/DependencyInjection/Compiler/SetTestClientPass.php
+++ b/DependencyInjection/Compiler/SetTestClientPass.php
@@ -10,12 +10,7 @@ class SetTestClientPass implements CompilerPassInterface
 {
     public function process(ContainerBuilder $container)
     {
-        // "query.max_query_count" is an array, it is only accessible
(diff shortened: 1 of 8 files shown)


alexislefebvre:
Too bad, I found this way to define parameters only today, one day too late for using this for the 1.4.0 version. :)

I think it's cleaner than the way I used in the previous PR.

lsmith77:
yeah .. I guess its a minor BC break if people rely on the parameters rather than the functionality .. but they shouldn't so I think its fine and we can put this into 1.4.1

PR #243, opened 2016-02-08
Refactor loadFixtureFiles() with 2 new functions

alexislefebvre:
Split loadFixtureFiles() functions in 3 distinct functions


alexislefebvre:
I made it in response to the warnings from Scrutinizer. Feel free to suggest other names to the functions if they are not relevant.

PR #253, opened 2016-03-07
Add paratest options to run tests faster

huitiemesens:
Here a quick script to add paratest support.

I squizz my current scripts into only one command, which in fact is the only call to make to run tests.
I didn't add test since i don t think travis has paratest feature, but if it does, i ll add some in a next commit.

Hope this will help

Issue #257, opened 2016-03-26
Database doesn't reload if a dependent fixture changes


cmfcmf:
My database didn't update when I changed the contents of a dependent fixture:

Test case:
```php
$this->loadFixtures([AFixture::class]);
```
AFixture:
```php
class AFixture implements DependentFixtureInterface {
// ...
    public function getDependencies()
    {
        return [DependencyFixture::class];
    }
}
```
config_test.yml
```yml
liip_functional_test:
    cache_sqlite_db: true
```
If you now alter the `DependencyFixture` class, the database is not updated when the test case is executed. (using Symfony 3 + latest version of the bundle).

alexislefebvre:
*If you now alter the `DependencyFixture` class*

How do you do this? I'm pretty sure that the cache don't check the data of the fixture.

alexislefebvre:
When I need to modify an existing fixture, I create a class that use `$temp = $this->getReference('MY_BASE_FIXTURE');` then I can update the fixture and persist it. And I load this fixture in my tests. With this method, each fixture has a known state and can be cached.

cmfcmf:
> How do you do this? I'm pretty sure that the cache don't check the data of the fixtures.

> When I need to modify an existing fixture, I create a class that use $temp = $this->getReference('MY_BASE_FIXTURE'); then I can update the fixture and persist it. And I load this fixture in my tests. With this method, each fixture has a known state and can be cached. Hope this helps.

I am unsure whether I understand your comment and question. 
The issue is that the `\Liip\FunctionalTestBundle\Test\WebTestCase::isBackupUpToDate` method is called with the class names passed to the `loadFixtures` method (-> `[AFixture::class]` in my example). 
The `isBackupUpToDate` method doesn't check whether fixtures defined in the `AFixture::getDependencies()` are up to date (-> `[DependencyFixture::class]` in my example) but only checks whether `AFixture::class` is up to date.

alexislefebvre:
What do you mean by *fixtures [...] are up to date*? How do you update the fixtures or its dependencies?

cmfcmf:
> What do you mean by fixtures [...] are up to date? How do you update the fixtures or its dependencies?

I change the code inside the `load` method of my `AFixture` or `DependencyFixture` classes.

alexislefebvre:
Thanks, I understand now.

[...]

PR #258, opened 2016-03-26
Fix tests from #253:

alexislefebvre:
Use runCommand() instead of StreamOutput and BufferedOutput
Get cache directory with getParameter()
Use $output->write() instead of echo()


alexislefebvre:
Hi @huitiemesens, can you please review this PR?

In `Command/RunParatestCommand.php`, cache directory is read with `getParameter()` and string concatenation is used instead of variable interpretation in string with double quotes.

In `Tests/Command/ParatestCommandTest.php`, tests about output are replaced by `runCommand()`.

Thanks.

huitiemesens:
Hi, i ll dot it on lunch time, sorry i was really busy today because of
work .

2016-03-26 20:00 GMT+01:00 Alexis Lefebvre <notifications@github.com>:

> Hi @huitiemesens <https://github.com/huitiemesens>, can you please review
> this PR?
>
> In Command/RunParatestCommand.php, cache directory is read with
> getParameter() and string concatenation is used instead of variable
> interpretation in string with double quotes.
>
> In Tests/Command/ParatestCommandTest.php, tests about output are replaced
> by runCommand().
>
> Thanks.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/liip/LiipFunctionalTestBundle/pull/258#issuecomment-201911840>
>



-- 

--

*Charles-Antoine FOURNEL*

Directeur Associé

+33 6 07 69 35 18
Skype : charles.fournel

[image: Description : cid:3420461422_1495647]

2, rue Meissonier

75017 Paris


This email and any files transmitted with it are confidential and intended
solely for the use of the individual or entity to whom they are addressed.
If you have received this email in error please notify the system manager.
This message contains confidential information and is intended only for
the individual named. If you are not the named addressee you should
not disseminate, distribute or copy this e-mail.

alexislefebvre:
There is no urgency, you can take your time. :)

(I removed the personal data in your email signature, it was displayed at the end of your previous comment)

alexislefebvre:
@huitiemesens do you think you can review this PR during this week? Thanks.

huitiemesens:
@alexislefebvre  from what i have read , it s seems okay to me ! Just a dumb question, but how can i pull this PR to test it ? I m a bitbucket guy and i m still not confortable with github mechanism

alexislefebvre:
Thanks.

In order to test this PR, I think that you have to clone my fork then checkout the corresponding branch: https://github.com/alexislefebvre/LiipFunctionalTestBundle/tree/253-fix-tests

This should work:

    git clone https://github.com/alexislefebvre/LiipFunctionalTestBundle.git
    cd LiipFunctionalTestBundle/
    git checkout 253-fix-tests

huitiemesens:
@alexislefebvre i ve try it inside one of our project which large amount of tests and the result fails : I think we should add a double check for paratest bin ( 1 check for ./bin/paratest and 1 for /vendor/bin/paratest ) since inside a symfony project the bin path is in root directory. That way the liiip test for paratest would still work. In order to fully test it i copied the paratest bin inside vendor/bin/ but now i m facing a new issue : 

No such file or directory in ROOT_DIR/vendor/liip/functional-test-bundle/Tests/App/bootstrap.php on line 5

the command should use project environment and if not use the liipbundle environment

alexislefebvre:
Thanks a lot for your report, I'll change the code today.

And I should not have used `__DIR__.'/../Tests/Test/'` in the *real* command, it should be used in Bundle's own tests only.

huitiemesens:
@alexislefebvre  yes i guess it was for a quick test so i remove it in my own tests to run all my project test , but still the same issue

alexislefebvre:
One problem is that `phpunit` and `paratest` binaries are located in `vendor/bin/` with Symfony 2 and `bin/` with Symfony 3. Maybe it would be simpler to provide the paths to `phpunit`, `paratest` and `phpunit.xml.dist` when launching the `test:run` command.

alexislefebvre:
@huitiemesens I have added a line in the configuration in order to define the path to `paratest`.

alexislefebvre:
I was able to run the tests with paratest by using the command `php bin/paratest -p 4 --phpunit bin/phpunit -c app/phpunit.xml.dist` in one Symfony2 project. It worked great.

If I understand correctly the previous PR, the role of the `Command/RunParatestCommand.php` file is to refresh the database. I think it's a lot of complication for an edge case. Maybe we should better add a note in the documentation to inform users that they have to use `php app/console cache:clear --env=test` after they change an entity or update a fixture.

huitiemesens:
@alexislefebvre exactly it was juste an help command for lazy people ( including me :) ) . The fact is, if you refresh a fixture or one of your test update / delete or insert a row in dbs, you ll have to clear cache then run commeqnd to populate the cp the db for each process created the run paratest, a lot of work for one command.

alexislefebvre:
> if you refresh a fixture

That's the problem reported in #257.

> one of your test update / delete or insert a row in dbs, you ll have to clear cache

That's strange, I never had too clear the cache, the backup made by this Bundle contain the original data. If some data is inserted during a test and the database is updated, it should not change the backup. The backup will be loaded instead of this altered databases during the following tests.

> cp the db for each process created then run paratest

This Bundle is supposed to create the SQLite dabase automatically. The fact that I can run my tests with `php bin/paratest -p 4 --phpunit bin/phpunit -c app/phpunit.xml.dist` seems to indicate that this command is not necessary. Am I missing something?

huitiemesens:
Does the bundle make a backup of the duplicated db too ? Because it seems the only first one is backup, so i process#2 alter db2.db, next run could not execute the same test if triggered again by process#2 , correct ?

alexislefebvre:
I'm not sure so I won't change this.

alexislefebvre:
At first I wanted to fix the tests, so I changed a little bit of the configuration. But this PR has grown too much. I'll try to make the smallest PR possible, publish the 1.6 version then see if we can improve the `test:run` command.

alexislefebvre:
The tests included in this Bundle work but when I test this PR in a real project it fails, a database is not created for each process.

Command/RunParatestCommand.php
```diff
@@ -25,37 +26,26 @@ class RunParatestCommand extends ContainerAwareCommand
...
+        $this->xmlConfig = $container->getParameter('liip_functional_test.paratest.xml_config');
 
-        $this->output->writeln('Initial schema created');
-        $populateProcess = new Process("php app/console doctrine:fixtures:load -n --fixtures $this->testDbPath/../src/overlord/AppBundle/Tests/DataFixtures/ORM/ --env=test");
-        $populateProcess->run();
+        $this->testDbPath = $container->getParameter('kernel.cache_dir');
 
-        $this->output->writeln('Initial schema populated, duplicating....');
-        for ($a = 0; $a < $this->process; ++$a) {
-            $test = new Process("cp $this->testDbPath/cache/test/dbTest.db ".$this->testDbPath."/cache/test/dbTest$a.db");
```
huitiemesens: that loop is needed in order to copy original db for each process

alexislefebvre: But LiipFunctionalTestBundle doesn't use any code like this when we don't use paratest, Doctrine creates the database automatically. I don't understand why it should be different with paratest. Isn't the database created automatically with the path provided in [`createConnection`](https://github.com/liip/LiipFunctionalTestBundle/pull/253/files#diff-83ff5a47a10ec9c63718ab50cad18db4R25)?

huitiemesens: The connectionFactory class i provide is just here to tell doctrine to use a corresponding DB according to process number , example : if process#3 run a test, this class will tell it to use dbtest3.db thanks to getenv('TEST_TOKEN'). That's why we need a process in the command  to multiply db with amount of process. When i start to use paratest at first, i wasn t using this class so all the phpunit process were using the same db, but i had I/O conflict or test using data previously deleted by another test ( like a test series to create / update / delete a user ) .

alexislefebvre:
I'm closing this in favor of the smaller PR #265.

PR #259, opened 2016-04-03
README.md: Add note about clearing cache when cache_sqlite_db is enabled

alexislefebvre:
Adresses #257.

### DIFF
--- a/README.md
+++ b/README.md
@@ -312,6 +312,9 @@ Tips for Fixture Loading Tests
         cache_sqlite_db: true
     ```
 
+    Note: If you use `DependentFixtureInterface` in your fixtures,
+    see the note in [Caveats](#caveats) about the caching of the SQLite database.
+
  3. Load your Doctrine fixtures in your tests:
 
     ```php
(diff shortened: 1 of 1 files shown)


alexislefebvre:
Can you please review this PR? ping @cmfcmf @huitiemesens Thanks.

README.md
```diff
@@ -312,6 +312,9 @@ Tips for Fixture Loading Tests
         cache_sqlite_db: true
     ```
 
+    Note: You have to clear the cache every time you modify an entity or a fixture:
```
cmfcmf: This is not true. It happens when you modify a fixture which is only referenced by the main fixture you pass to the `loadFixtures` method.

alexislefebvre: Thanks for you comment. I wanted to cover all the cases but this is too broad. Note that if you use several levels of dependent fixtures, a fixture is not necessarily referenced by the main fixture.

E.g.: A depends on B, B depends on C . If you modify C, LiipFunctionalTestBundle won't detect the change.

cmfcmf:
Looks good to me now.

alexislefebvre:
If I understand the code correctly, [`$om->getMetadataFactory()->getAllMetadata()`](https://github.com/liip/LiipFunctionalTestBundle/blob/617cca7a87ab2551dcaba025c22331168c8492ec/Test/WebTestCase.php#L387-L393) should change once an entity has been updated, and since it's used to calculate the hash with `md5`, it should generate a new backup file. So it should not be necessary to clear the cache every time an entity is updated.

PR #263, opened 2016-04-12
Set paratest defaults and fixed RunParatestCommand

mpclarkson:
Currently you cannot override the default `paratest` configuration. 

The following line in `RunParatestCommand.php` returns a `bool` not an `array`.

```php
$this->configuration = $this->getContainer()->hasParameter('liip_functional_test');
```

This means that the `isset($this->configuration['paratest'])` checks all fail and the defaults for `process` and `phpunit` are used. However, as these defaults are **hardcoded** in the command class, the configuration options are not used.

In this PR I have:

1. Updated the `RunParatestCommand.php` to use the config parameters and removed the hardcoded values. 
2. Set the default `process` to `5` and the default `phpunit` to `./bin/phpunit` in `DependencyInjection/Configuration.php`. Previously, they were just empty strings.


alexislefebvre:
Ping @huitiemesens

huitiemesens:
I ll have a test within our projects this afternoon ;)

huitiemesens:
@mpclarkson  @alexislefebvre  its works fine within a sf2.8 project ( 1367 assertions / 199 tests in 27 seconds) if i change the configuration file path from "-c App/phpunit.xml.dist" to "-c app/" .

I know LiipBundle tests are under the "App" path which differ from Sf app folder case

alexislefebvre:
@huitiemesens Thanks for the review! 

@mpclarkson Can you please add some tests of the default configuration in the following file? https://github.com/liip/LiipFunctionalTestBundle/blob/1562ff83f3030b4f8afae9c100ac0253be372e6d/Tests/DependencyInjection/ConfigurationTest.php Thanks

mpclarkson:
@alexislefebvre Done!

alexislefebvre:
Thanks!

PR #265, opened 2016-04-30
Fix test of the "paratest:run" command:

alexislefebvre:
- New: Pass optional arguments from "paratest:run" to the paratest command
- Define path to PHPUnit in AppConfig/config.yml
- Use runCommand() instead of StreamOutput and BufferedOutput in order to get the output


alexislefebvre:
I tested it in one of my projects and it work.

PR #267, opened 2016-05-22
Extract fixtures and assertions

alexislefebvre:
Use services instead of inheritance from WebTestCase


alexislefebvre:
This is a test for an hypothetical 2.0 version. This is not ready yet.

alexislefebvre:
I expected a difference between the current version and this version that use a lighter `WebTestCase` but the PHPUnit tests take the same time and the same amount of memory.

PR #271, opened 2016-06-12
Webtestcase: Extract some fixtures methods

alexislefebvre:
to the new FixturesLoader class

### DIFF
--- a/Test/WebTestCase.php
+++ b/Test/WebTestCase.php
@@ -311,9 +311,7 @@ abstract class WebTestCase extends BaseWebTestCase
         $om = $registry->getManager($omName);
         $type = $registry->getName();
 
-        $executorClass = 'PHPCR' === $type && class_exists('Doctrine\Bundle\PHPCRBundle\DataFixtures\PHPCRExecutor')
-            ? 'Doctrine\Bundle\PHPCRBundle\DataFixtures\PHPCRExecutor'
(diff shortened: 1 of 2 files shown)


alexislefebvre:
It would be better to move all the code related to fixtures in a new class  but it's impossible because it would break the callbacks calls. So this is a compromise.

lsmith77:
any specific reason why you made the implementation using static methods? limits the ability to extend if needed.

alexislefebvre:
No I had no reason to declare the methods as static. But since it calls functions directly like `FixturesLoader::locateResources(…);`, can `FixturesLoader` be extended?

lsmith77:
well if you add a factory method to the base class or make it an optional parameter then yes ..

alexislefebvre:
I remember now why I declared static methods: it was easier to use them since it doesn't require to instantiate an object before calling the methods from `FixturesLoader`.

> well if you add a factory method to the base class or make it an optional parameter then yes ..

I think that I understand your idea now.

Would adding a method like `protected function getFixturesLoader() {}` in `WebTestCase` that

1. return an instance of `FixturesLoader` or instantiate an object then return it
2. can be overridden by any child class

is a good solution?

lsmith77:
yeah .. I think so 

(sorry for the late reply)

alexislefebvre:
This should be done with traits: https://github.com/liip/LiipFunctionalTestBundle/issues/332#issuecomment-329958638

PR #278, opened 2016-08-28
Fixes #276: WebTestCase: Pass $omName to cleanDatabase()

### LINKED ISSUE #276: Clean Database doesn't work for EntityManagers other than default
Because of [this line](https://github.com/liip/LiipFunctionalTestBundle/blob/master/Test/WebTestCase.php#L487), the `cleanDatabase` call will always result in the default entity manager being cleaned. The `omName` Argument must be passed in the call to `loadFixtures()`.

alexislefebvre:
Fixes #276.

### DIFF
--- a/Test/WebTestCase.php
+++ b/Test/WebTestCase.php
@@ -472,8 +472,9 @@ abstract class WebTestCase extends BaseWebTestCase
      *
      * @param ManagerRegistry $registry
      * @param EntityManager   $om
+     * @param null            $omName
      */
-    private function cleanDatabase(ManagerRegistry $registry, EntityManager $om)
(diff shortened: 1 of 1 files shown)


Test/WebTestCase.php
```diff
@@ -545,7 +546,7 @@ public function loadFixtureFiles(array $paths = array(), $append = false, $omNam
         $om = $registry->getManager($omName);
 
         if ($append === false) {
-            $this->cleanDatabase($registry, $om);
+            $this->cleanDatabase($registry, $om, $omName);
```
alexislefebvre: I don't know how to retrieve the name `$omName` from `$om`, so I took the argument `$omName` directly.

PR #293, opened 2016-12-14
fix deprecation warnings with symfony 2.7

lsmith77:


Tests/App/bootstrap.php
```diff
@@ -1,9 +1,12 @@
 <?php
 
 use Doctrine\Common\Annotations\AnnotationRegistry;
+use Symfony\Bridge\PhpUnit\DeprecationErrorHandler;
 
 $loader = require __DIR__.'/../../vendor/autoload.php';
 
 AnnotationRegistry::registerLoader(array($loader, 'loadClass'));
 
+DeprecationErrorHandler::register(getenv('SYMFONY_DEPRECATIONS_HELPER'));
```
stof: you don't need it. The deprecation error handler is registered automatically when calling phpunit and the bridge is installed through composer

stof:
to fix the deprecations, start by using Doctrine ORM 2.5 rather than 2.4 (2.4 is EOLed and so has not migrated away from deprecated APIs)

lsmith77:
sigh https://travis-ci.org/liip/LiipFunctionalTestBundle/jobs/184549050 ..
need to look into that to see if we need to bump some more lower versions

stof:
@lsmith77 the issue is that it installs both symfony/symfony 2.3 and symfony/console 2.5 (due to an long-time issue in composer where an optimization broke the handling of conflicts for replaced versions). And as you rely on detecting the kernel version to decide whether the console component supports a feature or no, this breaks things when versions are different.
Btw, the same would be true if the fullstack repo was not installed at all.
The fix for that is simply: use feature detection instead (similar to the rule about doing feature detection rather than browser sniffing in JS btw)

.travis.yml
```diff
@@ -18,22 +18,24 @@ services:
 
 matrix:
   include:
-    - php: 5.3
+    - php: 5.4
       env: COMPOSER_FLAGS="--prefer-lowest"
```
alexislefebvre: You can move this section in the [allowed failures](https://docs.travis-ci.com/user/customizing-the-build#Rows-that-are-Allowed-to-Fail), `--prefer-lowest` install outdated versions and waste time during PRs.

lsmith77: @alexislefebvre my goal is to make sure that we simply bump the lowest version to a sane point .. f.e. I just bumped the minimal version to 2.3.2 since it makes no sense to maintain a 2.3.1 specific hack

alexislefebvre: @lsmith77 You're right, I misunderstood the `--prefer-lowest` flag.

lsmith77: hmm .. decided to revert removing the hack .. but now ``--prefer-lowest`` is passing and everything else is failing .. but this looks fixable .. will try to figure it out over the weekend. but if someone has some time to kill I wouldn't mind a PR :)

lsmith77:
@stof I am trying to find the relevant change but not really finding it https://github.com/symfony/console/compare/v2.3.0...v2.5.0

also its locally with `--prefer-lowest` I get 2.3.27 and on travis 2.3.42 .. which is why I tried the `composer self-update`.

PR #296, opened 2016-12-21
Fix conflict of in-memory sqlite & disabled cache_sqlite_db param

AlexMakII:
Make WebTestCase check sqlite path only if backup is enabled.
Should fix #84. At least removes weird error.


AlexMakII:
Its a fast fix. So it's quite dirty.
Alternative way is:
1. adding $in_memory flag and proceed corresponding everywhere it's suitable
2. disable backups with warnings or throw exception

AlexMakII:
Sorry, was a mistake

PR #299, opened 2017-01-04
Handle Symfony disabling the validator component

### LINKED ISSUE #288: Bundle should not require Validator to be enabled
The Bundle defines a [decorator on the `validator` service](https://github.com/liip/LiipFunctionalTestBundle/blob/master/Resources/config/validator.xml). Currently [this is only loaded if the `ValidatorInterface` exists](https://github.com/liip/LiipFunctionalTestBundle/blob/9ce4e7f1948e559d986141b57f913bb83fcc9734/DependencyInjection/LiipFunctionalTestExtension.php#L33). 

However it is possible for the `ValidatorInterface` to exist but the `validator` service to not be available if the user has not [enabled validation within the FrameworkBundle configuration](http://symfony.com/doc/current/reference/configuration/framework.html#reference-validation-enabled).

This occurs in my case because I have been attempting to streamline my DI container by disable unused services, and I have disabled validation as I do no require it within my code. However now my tests fail with the message `Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException: You have requested a non-existent service "validator".` when attempting to build the container for my tests because the `LiipFunctionalTestBundle` config references this non-existent service.

---

LiipFunctionalTestBundle should be able to create a valid container even when the `validator` component is not present in the DI system.

It seems like I can fix this by adding the following in a compiler pass, but I'm unsure if this is the best solution as not including the config in the first place sounds preferable.

```php
if (!$container->hasDefinition('validator')) {
    $container->removeDefinition('liip_functional_test.validator');
}
```

In addition to this, references to `liip_functional_test.validator` within [HttpAssertions](https://github.com/liip/LiipFunctionalTestBundle/blob/9ce4e7f1948e559d986141b57f913bb83fcc9734/Utils/HttpAssertions.php) should include a `getContainer()->has()` check to assert on that the service exists before attempting to `get()` it. This would also have the side-effect of allowing Symfony <2.5 users to use assertStatusCode() in their tests.

BPScott:
Symfony now allows the user to disable the validator component, which
results in the validator service being undefined. This then causes
errors on container compilation as we expected this to always exist.

To fix this add a compiler pass that removes our wrapper around
validator if the validator itself is undefined.

Fixes #288.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,8 @@ vendor
 /Tests/App/logs/
 /Tests/AppConfig/cache/
 /Tests/AppConfig/logs/
+/Tests/AppConfigLeanFramework/cache/
+/Tests/AppConfigLeanFramework/logs/
 /Tests/AppConfigMysql/cache/
 /Tests/AppConfigMysql/logs/
 /Tests/AppConfigPhpcr/cache/
--- a/DependencyInjection/Compiler/OptionalValidatorPass.php
+++ b/DependencyInjection/Compiler/OptionalValidatorPass.php
@@ -1,0 +1,16 @@
+<?php
+
+namespace Liip\FunctionalTestBundle\DependencyInjection\Compiler;
+
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+
+class OptionalValidatorPass implements CompilerPassInterface
+{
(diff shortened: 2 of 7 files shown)


lsmith77:
thx

lsmith77:
@alexislefebvre I guess its time to make a 1.7.0 release .. wdyt?

alexislefebvre:
@lsmith77 :+1: for a new release. :)

PR #300, opened 2017-01-17
Add purge mode for to load fixtures from files.

sane4ek-2:
Add purge mode for to load fixtures from files because need to truncate tables in some situations
and method of load fixture doesn't allow to switch off FOREIGN_KEY_CHECKS.

### DIFF
--- a/README.md
+++ b/README.md
@@ -415,6 +415,37 @@ $fixtures = $this->loadFixtureFiles(array(
 ));
 ```
 
+If you want to clear tables you have the two ways:
+1. Only to remove records of tables;
+2. Truncate tables.
+
+The first way is consisted in using the second parameter `$append` with value `true`. It allows you **only** to remove all records of table. Values of auto increment won't be reset. 
+```php
+$fixtures = $this->loadFixtureFiles(
+    array(
+        '@AcmeBundle/DataFixtures/ORM/ObjectData.yml',
+        '@AcmeBundle/DataFixtures/ORM/AnotherObjectData.yml',
+        '../../DataFixtures/ORM/YetAnotherObjectData.yml',
+    ),
+    true
+);
+```
+
+The second way is consisted in using the second parameter `$append` with value `true` and the last parameter `$purgeMode` with value `Doctrine\Common\DataFixtures\Purger\ORMPurger::PURGE_MODE_TRUNCATE`. It allows you to remove all records of tables with resetting value of auto increment.
+
+```php
+<?php
+
+use Doctrine\Common\DataFixtures\Purger\ORMPurger;
+
+$files = array(
+     '@AcmeBundle/DataFixtures/ORM/ObjectData.yml',
+     '@AcmeBundle/DataFixtures/ORM/AnotherObjectData.yml',
+     '../../DataFixtures/ORM/YetAnotherObjectData.yml',
+ );
+$fixtures = $this->loadFixtureFiles($files, true, null, 'doctrine', ORMPurger::PURGE_MODE_TRUNCATE );
+```
+
 #### HautelookAliceBundle Faker Providers
 
 This bundle supports faker providers from HautelookAliceBundle.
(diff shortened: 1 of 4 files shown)


alexislefebvre:
How do you use this new parameter? Something like `loadFixtureFiles([…], ORMPurger::PURGE_MODE_DELETE)`? Do values come from [these constants](https://github.com/doctrine/data-fixtures/blob/fbd7abf58f7496998d79e37087554f8dbd6ae7d2/lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php#L36-L37)? Thanks.

sane4ek-2:
Yes, you are right.

alexislefebvre:
Can you please add an explanation in the [documentation](https://github.com/liip/LiipFunctionalTestBundle#loading-fixtures-using-alice)? While I'm writing this, I realize that none of the parameters is documented… Anyway, it would be awesome if you could add a line explaining the problem you encountered and how this parameter has to be changed. Thanks.

sane4ek-2:
Ok.

lsmith77:
docs are still missing.
also not sure if there is a risk of BC issues with the new parameter but I guess it has a default value so it should be fine. should just be confirmed before merging.

sane4ek-2:
I welcome everybody. I've added part of documentation. Sorry me that I was so lazy and didn't it so long.

lsmith77:
@alexislefebvre do you have time to review? I am not how soon I will have time to review it properly.

README.md
```diff
@@ -415,6 +415,37 @@ $fixtures = $this->loadFixtureFiles(array(
...
+<?php
+
+use Doctrine\Common\DataFixtures\Purger\ORMPurger;
+
+$files = array(
+     '@AcmeBundle/DataFixtures/ORM/ObjectData.yml',
+     '@AcmeBundle/DataFixtures/ORM/AnotherObjectData.yml',
+     '../../DataFixtures/ORM/YetAnotherObjectData.yml',
+ );
+$fixtures = $this->loadFixtureFiles($files, true, null, 'doctrine', ORMPurger::PURGE_MODE_TRUNCATE );
```
alexislefebvre: Can you please add a test by using this call in a test similar to [`testLoadFixturesFiles()`](https://github.com/liip/LiipFunctionalTestBundle/blob/537f220460c8af9571ea7030c588bbfba269e2e2/Tests/Test/WebTestCaseTest.php#L365-L414)?

I'm thinking to something like this:

```php
    /**
     * Use nelmio/alice with PURGE_MODE_TRUNCATE.
     */
    public function testLoadFixturesFilesWithPurgeModeTruncate()
    {
        $fixtures = $this->loadFixtureFiles(array(
            '@LiipFunctionalTestBundle/Tests/App/DataFixtures/ORM/user.yml',
        ), true, null, 'doctrine', ORMPurger::PURGE_MODE_TRUNCATE );

        $this->assertInternalType(
            'array',
            $fixtures
        );

        // 10 users are loaded
        $this->assertCount(
            10,
            $fixtures
        );
    }
```

It won't check that purge worked but at least it will ensure that using the new parameter does not break fixtures loading.

sane4ek-2: Ok. I'll try to add a test at this weekend.

sane4ek-2: The test `testLoadFixturesFilesWithPurgeModeTruncate` was added with a bit modifications. Also, as far as you know travis-ci gets error about HHVM on Ubuntu Precise and that test has failed.

lsmith77: feel free to drop HHVM support

sane4ek-2: How can I refresh result of tests? Do I need to make some commit into repo?

alexislefebvre: In order to remove HHVM, you have to delete [this line](https://github.com/liip/LiipFunctionalTestBundle/blob/165268dfee486012d5641698ebe9e4bddf7574ec/.travis.yml#L7) and push a commit on your branch.

Issue #301, opened 2017-01-20
How to load fixture yml files which are outside a bundle? 


Kwadz:
To load the fixtures, I use:

```
$fixtures = $this->loadFixtureFiles(array(
    '../../Resources/fixtures/orm/processor.yml',
));
```

and I get: `A resource name must start with @ ("../../DataFixtures/ORM/processor.yml" given).`

So, how to load yml files which are outside a bundle? 
In this case where should be located the `DataFixtures/ORM/` directory?

alexislefebvre:
It looks like the error message does not come from LiipFunctionalTestBundle. Can you please give the complete error backtrace?

Kwadz:
To reproduce the error:

1. Install new Symfony project
2. Install LiipFunctionalTestBundle: `composer require --dev liip/functional-test-bundle`
3. Install AliceBundle: `composer require --dev hautelook/alice-bundle:~1.2`
4. Install DoctrineFixturesBundle: `composer require --dev doctrine/doctrine-fixtures-bundle`
5. Add the following code:

```
use Liip\FunctionalTestBundle\Test\WebTestCase;

class DefaultControllerTest extends WebTestCase
{
    public function testLiip() {
        $fixtures = $this->loadFixtureFiles(array(
            '../../DataFixtures/ORM/YetAnotherObjectData.yml',
        ));
    }
}
``` 

Stacktrace:

```
A resource name must start with @ ("../../DataFixtures/ORM/YetAnotherObjectData.yml" given).
 /Users/richard/dev/my_sf3_project/vendor/symfony/symfony/src/Symfony/Component/HttpKernel/Kernel.php:213
 /Users/richard/dev/my_sf3_project/vendor/liip/functional-test-bundle/Test/WebTestCase.php:524
 /Users/richard/dev/my_sf3_project/vendor/liip/functional-test-bundle/Test/WebTestCase.php:562
 /Users/richard/dev/my_sf3_project/tests/AppBundle/Controller/DefaultControllerTest.php:22
```

alexislefebvre:
Thanks. The error comes from these lines: https://github.com/liip/LiipFunctionalTestBundle/blob/dc7eb1b2fbb2f2c508c121f7ee0bb4ed1a96128b/Test/WebTestCase.php#L518-L525

[...]

PR #302, opened 2017-01-21
Add support for wrong fixture path outside a bundle

Kwadz:
Related to #301

### DIFF
--- a/Test/WebTestCase.php
+++ b/Test/WebTestCase.php
@@ -520,7 +520,10 @@ abstract class WebTestCase extends BaseWebTestCase
                 $files[] = $path;
                 continue;
             }
-
+            if ($path[0] !== '@' && !file_exists($path) === true) {
+                throw new \InvalidArgumentException(sprintf('Unable to find file "%s".', $path));
+            }
+            
             $files[] = $kernel->locateResource($path);
         }
 
--- a/Test/WebTestCase.php
+++ b/Test/WebTestCase.php
@@ -518,12 +518,12 @@ abstract class WebTestCase extends BaseWebTestCase
         $kernel = $this->getContainer()->get('kernel');
 
         foreach ($paths as $path) {
-            if ($path[0] !== '@' && file_exists($path) === true) {
+            if ($path[0] !== '@') {
+                if (!file_exists($path)) {
+                    throw new \InvalidArgumentException(sprintf('Unable to find file "%s".', $path));
+                }
                 $files[] = $path;
                 continue;
-            }
-            if ($path[0] !== '@' && !file_exists($path) === true) {
-                throw new \InvalidArgumentException(sprintf('Unable to find file "%s".', $path));
             }
 
             $files[] = $kernel->locateResource($path);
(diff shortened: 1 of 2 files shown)


Test/WebTestCase.php
```diff
@@ -520,6 +522,9 @@ private function locateResources($paths)
                 $files[] = $path;
                 continue;
             }
+            if ($path[0] !== '@' && !file_exists($path) === true) {
```
alexislefebvre: Is there a reason you didn't put this in an `else` block?

Kwadz:
I wanted to avoid nested `if` statements and keep parameters symmetric. But I agree with your code, in this case it doesn't make more sense having every parameters written 2 times.

Kwadz:
So, we go for the following?
```php
if ($path[0] !== '@') {
    if (!file_exists($path)) {
        throw new \InvalidArgumentException(sprintf('Unable to find file "%s".', $path));
    }
    $files[] = $path;
    continue;
}
```

alexislefebvre:
Sorry for thinking to this so late but, could you please add a corresponding test? Otherwise, I'll add a test by myself, that's not a problem. Thanks.

alexislefebvre:
Thanks for the test! I'm sorry to bother you, but can you please add a test that load successfully a fixture without the `@` at the beginning ?

I'm thinking to something like this:

```php
    /**
     * Use nelmio/alice.
     */
    public function testLoadFixturesFilesWithPathOutsideABundle()
    {
        $fixtures = $this->loadFixtureFiles(array(
            '[…]/Tests/App/DataFixtures/ORM/user.yml',
        ));
        $this->assertInternalType(
            'array',
            $fixtures
        );
        // 10 users are loaded
        $this->assertCount(
            10,
            $fixtures
        );
    }
```

Feel free to to add a fixture in another directory if this is required by the test.

See this test for reference: https://github.com/liip/LiipFunctionalTestBundle/blob/97cedfc1edd51e142044f2db3759bd96ef7a9d5a/Tests/Test/WebTestCaseTest.php#L365-L425 (I don't think that we perform all the tests, counting the number of loaded fixtures is sufficient)

Kwadz:
The `testLoadFixturesFilesPaths()` method already do the job. The returned value by Symfony Kernel is an absolute path.

https://github.com/liip/LiipFunctionalTestBundle/blob/master/Tests/Test/WebTestCaseTest.php#L422

alexislefebvre:
Thanks for your answer.

Yes, `testLoadFixturesFilesPaths()` will load the fixtures, but the test uses `$this->client->getContainer()->get('kernel')->locateResource(` with a `@` in the path, so I think that it is different from your original problem.

I added a test that doesn't rely on `locateResource()` and a `@…` fixture.

Kwadz:
In fact `$this->client->getContainer()->get('kernel')->locateResource()` is only used for test purpose. It returns the desired absolute path which will be used by `WebTestCase::loadFixtureFiles()` and therefore by `WebTestCase::locateResources()`. The goal is not to test Symfony Kernel in our case but `WebTestCase::locateResources()`. I think there is no need for additional test.

alexislefebvre:
Yes, this test was not necessary, I just wanted to ensure that the behaviour described in your issue #301 worked.

PR #305, opened 2017-02-19
Remove outdated code in Html5WebTestCase

### LINKED ISSUE #251: Remove old code with comment "this stuff should go away"?
I found a surprising comment in `Test/Html5WebTestCase.php` ([example in 1.5 version](https://github.com/liip/LiipFunctionalTestBundle/blob/1.5.0/Test/Html5WebTestCase.php#L156-L160)):

```
/*
 * unfortunately, the bamboo html5 validator.nu gives back an empty "message" about the error with brightcove object, but we have to ignore the error
 * if our local validator.nu instance is fixed, this stuff should go away
 */
$ignores_extract = $this->getContainer()->getParameter('liip_functional_test.html5validation.ignores_extract');
```

Should this code be removed?

It was added 5 years ago: https://github.com/liip/LiipFunctionalTestBundle/commit/a4a919039e07df073c73d4f13eb8e12299e2d824

alexislefebvre:
Fixes #251


alexislefebvre:
We probably shouldn't remove this before LiipFunctionalTestBundle 2.0, because it is a backward compatibility (even if this feature may have been not used much).

PR #308, opened 2017-03-10
Support for running PHPUnit from outside the test case class directory

Kwadz:
Currently `phpunit` command can be executed only from the directory of the test case class, else we get the following exception while loading fixtures with relative path:
`PHP Fatal error:  Uncaught InvalidArgumentException: Unable to find file "../../Resources/fixtures/orm/my_fixture.yml"`

For example, in other words, we can't execute `phpunit` from the project root directory.

This commit allows to execute `phpunit` from everywhere.

### DIFF
--- a/Test/WebTestCase.php
+++ b/Test/WebTestCase.php
@@ -529,20 +529,24 @@ abstract class WebTestCase extends BaseWebTestCase
         return $files;
     }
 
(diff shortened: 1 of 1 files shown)


Kwadz:
Fix #309

Kwadz:
Closed as fixed by #333

PR #310, opened 2017-03-20
Allow loadFixtures() without arguments

Grafikart:


alexislefebvre:
`$this->loadFixtures(array());` already work for loading empty fixtures. With you PR we can call `$this->loadFixtures();`? I'm asking this because the title of this PR is a little bit confusing. :)

Can you please add a corresponding test? You just have to create a function similar to [`testLoadEmptyFixtures()`](https://github.com/liip/LiipFunctionalTestBundle/blob/09b6f60869ef84f642d110c465611521f23a9da8/Tests/Test/WebTestCaseTest.php#L270-L281), but without `array()`. Thanks

Grafikart:
Yeah sorry the title is indeed misleading. I did the PR using github directly, I'll create a branch to add the test later this afternoon

Grafikart:
Do I create a new test or change the current one ?

alexislefebvre:
Please add a new test (a new function), so we'll check `$this->loadFixtures(array());` and `$this->loadFixtures();` both work.

Grafikart:
I just added the test

alexislefebvre:
Thanks!

alexislefebvre:
Hi @Grafikart, it looks like we will have to revert this change because it breaks some tests environment: see #311. I don't see another solution.

PR #315, opened 2017-04-11
Validator decorator looses a root object in context

mrix:
Hi. There is a bug when the validator wrapper looses a `root` object in the context.
For example, this code fails only when this bundle is enabled:

```php
class MyCustomConstraintValidator extends ConstraintValidator
{
    /**
     * @inheritdoc
     */
    public function validate($value, Constraint $constraint)
    {
        $root = $this->context->getRoot();
        if ($root === null) {
            throw new UnexpectedTypeException($root, 'not null');
        }
    }
}
```

The problem is in [this method](https://github.com/symfony/symfony/blob/v2.7.26/src/Symfony/Component/Validator/Validator/RecursiveValidator.php#L74): `RecursiveValidator.startContext()` does not match the interface signature.

### DIFF
--- a/Validator/DataCollectingValidator.php
+++ b/Validator/DataCollectingValidator.php
@@ -71,9 +71,9 @@ class DataCollectingValidator implements ValidatorInterface, EventSubscriberInte
         return $this->wrappedValidator->validatePropertyValue($objectOrClass, $propertyName, $value, $groups);
     }
 
-    public function startContext()
+    public function startContext($root = null)
     {
-        return $this->wrappedValidator->startContext();
+        return $this->wrappedValidator->startContext($root);
     }
 
     public function inContext(ExecutionContextInterface $context)


mrix:
BTW I'd like to add a new functional test for this bug but I don't understand how to do that.
Which application in the `Tests` directory should I modify?

alexislefebvre:
I'm not sure to understand the problem but you can see tests with validation here: https://github.com/liip/LiipFunctionalTestBundle/blob/537f220460c8af9571ea7030c588bbfba269e2e2/Tests/Test/WebTestCaseTest.php#L611-L636

mrix:
I meant that Symfony validator allows validating values in a different context - something like this:
```php
$errors = $this->get('validator')
    ->startContext(new SomeValidationContextRoot('some data...'))
    ->validate($object)
    ->getViolations();
```

But I've already found a solution without modifying this bundle.

PR #316, opened 2017-04-11
Suggest to enable bundle only in test environment

garak:
No need to enable bundle in dev

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,8 +54,11 @@ Installation
         public function registerBundles()
         {
             // ...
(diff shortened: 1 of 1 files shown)


alexislefebvre:
It looks good, thanks.

Did you have any problem when it was loading in the `dev` environment?

garak:
No problems so far, but some time ago Standard Edition started to suggest more specific environment activation and I like to follow such suggestions :wink:

PR #317, opened 2017-04-18
Adds backup file param even to preFixtureRestore and postFixtureRestore

mmenozzi:
Backup file param was passed only to `postReferenceSave` and `preReferenceSave` callbacks but not to `preFixtureRestore` and `postFixtureRestore`. Implementors may need this parameter in those callbacks.

### DIFF
--- a/Test/WebTestCase.php
+++ b/Test/WebTestCase.php
@@ -407,7 +407,7 @@ abstract class WebTestCase extends BaseWebTestCase
                         $om->flush();
                         $om->clear();
 
-                        $this->preFixtureRestore($om, $referenceRepository);
+                        $this->preFixtureRestore($om, $referenceRepository, $backup);
 
                         copy($backup, $name);
 
@@ -415,7 +415,7 @@ abstract class WebTestCase extends BaseWebTestCase
                         $executor->setReferenceRepository($referenceRepository);
                         $executor->getReferenceRepository()->load($backup);
 
-                        $this->postFixtureRestore();
+                        $this->postFixtureRestore($backup);
 
                         return $executor;
                     }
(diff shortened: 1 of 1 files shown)


alexislefebvre:
The last time I merged a similar PR, I had to revert it because it broke some methods that required the same signature (see #312). This is a Backward Compatibility break, right?

mmenozzi:
I think that it depends on php strictness. Anyway do you agree with this pr? If so I can see if there's some way to do it without bc break.

alexislefebvre:
Can you please give an use case? Why does someone may need to access to the backup path?

mmenozzi:
I work on a project where fixtures loads objects which needs related files on file system (for example a User object with profile picture on file system). On that project I also use a virtual file system library (https://github.com/mikey179/vfsStream) and when I load a User fixture I also need related file on the virtual file system and this is done by the fixture class itself. But with the sqlite db cache enabled when there's a cache hit those files aren't created on vfs because the fixture load method isn't executed.
So I implemented a vfs cache mechanism similar to the database one. In the preReferenceSave method I save the vfs structure serialized on a file whose path is the db backup file path plus the ".vfs" extension; then in the preFixtureRestore I need the same backup file path so I can check if there is a "$backup.vfs" file and, if so, unserialize the structure and load it on the virtual file system.

mmenozzi:
Hi @alexislefebvre, the thumbs up reaction means that you approve this change?
For the BC break I think that we could mark the `preFixtureRestore` and `postFixtureRestore` as deprecated and add two additional methods `preFixtureBackupRestore` and `postFixtureBackupRestore` with the additional parameter. I think that these new names are better and at the same time we can add the new parameter.

What do you think?

alexislefebvre:
The thumb up was a way to thank you for the detailed explanation. :)

I like the idea of adding a method with an additional parameter. What do you think @lsmith77?

lsmith77:
sounds like a good plan to me

mmenozzi:
@alexislefebvre @lsmith77 here we go, PR updated! Sounds good?
If yes, when you plan to release the versione 1.8.0 with this new feature?

lsmith77:
didn't have any specific plans .. but sure lets look towards a 1.8.0 release ASAP:
https://github.com/liip/LiipFunctionalTestBundle/milestone/3

mmenozzi:
@lsmith77 ok thanks!

Issue #323, opened 2017-06-01
Move all files to `src` and `tests`


ruudk:
Is it ok if I create a PR that moves all sources to `src` and the tests in `tests`? This way, the tests won't get distributed with composer.

I'm currently trying to migrate my Symfony 3.2 project to the Symfony 3.0 directory structure format and I'm getting weird errors like this:
```
$ composer install
...
...\Core\Composer\ScriptHandler::clearCache
PHP Fatal error:  Cannot declare class AppKernel, because the name is already in use in /vagrant/vendor/liip/functional-test-bundle/Tests/App/AppKernel.php on line 54

Fatal error: Cannot declare class AppKernel, because the name is already in use in /vagrant/vendor/liip/functional-test-bundle/Tests/App/AppKernel.php on line 54
Segmentation fault (core dumped)
Script ....\Core\Composer\ScriptHandler::clearCache handling the post-install-cmd event terminated with an exception


  [RuntimeException]
  An error occurred when executing the "'cache:clear --no-warmup'" command:
  Fatal error: Cannot declare class AppKernel, because the name is already in use in /vagrant/vendor/liip
  /functional-test-bundle/Tests/App/AppKernel.php on line 54
  PHP Fatal error:  Cannot declare class AppKernel, because the name is already in use in /vagrant/vendor
  /liip/functional-test-bundle/Tests/App/AppKernel.php on line 54
  Segmentation fault (core dumped)


install [--prefer-source] [--prefer-dist] [--dry-run] [--dev] [--no-dev] [--no-custom-installers] [--no-autoloader] [--no-scripts] [--no-progress] [--no-suggest] [-v|vv|vvv|--verbose] [-o|--optimize-autoloader] [-a|--classmap-authoritative] [--apcu-autoloader] [--ignore-platform-reqs] [--] [<packages>]...
```

alexislefebvre:
It will probably be simpler to exclude some directories from the archive: https://stackoverflow.com/questions/17049313/how-to-ignore-directories-with-composer

ruudk:
@alexislefebvre Thanks! I found that one as well, wasn't sure if it would work. Going to test it now in https://github.com/liip/LiipFunctionalTestBundle/pull/325

ruudk:
@alexislefebvre It doesn't work. Still gives the same error. So I think it's better to just split it.

alexislefebvre:
[...]

PR #324, opened 2017-06-01
Move files to `src` and `tests` directory

ruudk:
See #323 

Moving all the sources to `src` and tests to `tests`.

This way, the tests won't be distributed when people use this library.

Tests are failing because of HHVM #326


alexislefebvre:
Is it a feature from Composer to not share files in the `tests` directory?

ruudk:
Yep, see this in composer.json
```
 "autoload": {
        "psr-4": {
            "Liip\\FunctionalTestBundle\\": "src/"
        }
    },
    "autoload-dev": {
        "psr-4": {
            "Liip\\FunctionalTestBundle\\Tests\\": "tests/"
        }
    },
```

ruudk:
It makes sure that when I require this is a normal dependency (non dev) it will only pull `src`

PR #325, opened 2017-06-01
Composer > Exclude `Tests` directory from classmap

ruudk:
Trying a different approach for #323

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -50,7 +50,8 @@
     "autoload": {
         "psr-4": {
             "Liip\\FunctionalTestBundle\\": ""
-        }
+        },
+        "exclude-from-classmap": ["/Tests/"]
     },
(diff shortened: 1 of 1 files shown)


ruudk:
Doesn't work. Still gives 
```
  [RuntimeException]
  An error occurred when executing the "'cache:clear --no-warmup'" command:
  Fatal error: Cannot declare class AppKernel, because the name is already in use in /vagrant/vendor/liip
  /functional-test-bundle/Tests/App/AppKernel.php on line 54
  PHP Fatal error:  Cannot declare class AppKernel, because the name is already in use in /vagrant/vendor
  /liip/functional-test-bundle/Tests/App/AppKernel.php on line 54
  Segmentation fault (core dumped)
```

Issue #326, opened 2017-06-01
Remove HHVM support?


ruudk:
Tests fail on Travis:
> HHVM is no longer supported on Ubuntu Precise. Please consider using Trusty with `dist: trusty`.

See https://travis-ci.org/liip/LiipFunctionalTestBundle/jobs/238420775

alexislefebvre:
Is HHVM supported on Travis CI's Trusty?

ruudk:
Seems like it. Do we want to switch to that?

alexislefebvre:
I would like to wait the merge of the ongoing PR #322 before changing the test suite. Or the test suite should be uploaded in this PR. I don't know yet.

Soullivaneuh:
HHVM support is dropped from Symfony, so it would make sense to drop it here.

alexislefebvre:
We already removed it from tests: https://github.com/liip/LiipFunctionalTestBundle/commit/082a2134a1c2d3fc3a9a75952cca8e4e5ddf80c0#diff-354f30a63fb0907d4ad57269548329e3

So I'm closing it.

PR #327, opened 2017-06-02
Travis CI: Use Ubuntu Trusty

alexislefebvre:
See #326.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,5 @@
+dist: trusty
+
 language: php
 
 php:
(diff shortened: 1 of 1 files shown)


alexislefebvre:
There was a problem while creating the container:

> sudo service mysql start
> start: Job failed to start

z38:
The failure might be related to travis-ci/travis-ci#7790 which mentions the same issues when switching to Trusty.

alexislefebvre:
Thanks for the information @z38. We'll have to wait for the MySQL bug to be fixed before upgrading the Ubuntu version onTravis CI.

alexislefebvre:
Thanks for the information @z38. We'll have to wait for the MySQL bug to be fixed before upgrading the Ubuntu version on Travis CI.

z38:
According to the reports in travis-ci/travis-ci#8137, the bug seems to be fixed. So in theory, the version upgrade should work now.

alexislefebvre:
Thanks for the updates @z38! I'm merging since it works now.

PR #330, opened 2017-08-31
Enable application manipulation for command helper

dudologicus:
Currently the application is instantiated within the command helper method. This doesn't allow any manipulation (and configuration is not given at the moment).

For test cases I want to avoid that the command catches exceptions (e.g. assertion fails in callbacks).

Minimal invasive way (IMHO) is to allow to override Application instantiation.

```

class MyCommand() extends ContainerAwareCommand
{
    //...

    public function execute()
    {
        $myService = $this->getContainer()->get('my.service');
        $myService->doSomething();
    }
}

class MyTestCase extends WebTestCase
{
    //...

    public function testSomeSpecialCase()
    {

        $mock = $this->getMockBuilder(MyService::class)
                ->disableOriginalConstructor()
                ->setMethods(['doSomething'])
                ->getMock();

        $mock->expects(self::any())
            ->method('doSomething')
            ->will(
                self::returnCallback(
                    function () {
                        TestCase::assertEquals(1, 2);
                    }
                )
            );

        $this->getContainer()->set('my.service', $mock);

        $this->runCommand('my.command', [], true);
    }

    protected function createApplication(KernelInterface $kernel)
    {
        $application = parent::createApplication($kernel);
        $application->setCatchExceptions(false);
        return $application
    }
}
```

without setting `setCatchExceptions(false)`, this test would pass and render an exception in the output of the command.

### DIFF
--- a/Test/WebTestCase.php
+++ b/Test/WebTestCase.php
@@ -151,6 +151,7 @@ abstract class WebTestCase extends BaseWebTestCase
 
     /**
      * @param KernelInterface $kernel
+     *
      * @return Application
      */
     protected function createApplication(KernelInterface $kernel)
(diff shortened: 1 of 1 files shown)


alexislefebvre:
@dudologicus Could you please fix the StyleCI error?

PR #333, opened 2017-09-16
Update the fixtures path in the documentation

### LINKED ISSUE #309: Can't execute phpunit outside the test case class when using yml fixtures
Currently `phpunit` can be executed from the directory of the test case class only, else we get the following exception while loading fixtures with relative path:

`PHP Fatal error: Uncaught InvalidArgumentException: Unable to find file "../../Resources/fixtures/orm/my_fixture.yml"`

Kwadz:
This allows to execute PHPUnit from everywhere in the project, not only from the test case class directory.

Fix #309


alexislefebvre:
Thanks!

PR #321, opened 2017-05-15
#297 - added workaround for nelmio/alice v3

zhil:

PR #337, opened 2017-10-31
Nelmio Alice v3 upgrade

### LINKED ISSUE #268: nelmio/alice v3
I apologize for the spammy issue, but as LiipFunctionalTestBundle is a one of the consumer of [alice](https://github.com/nelmio/alice), I though worth informing you that the next version [v3 is on it's way](https://github.com/nelmio/alice/issues/265). So if you noticed something you would like to change at alice level, now is a good time :)

### LINKED ISSUE #297: WebTest case is not working with nelmio alice 3.0
A class "Fixtures" is need by the method loadFixtures, Isuppose it has become FixtureSet in nelmio\alice 3.0.

I get this error:
"BadMethodCallException: nelmio/alice should be installed to use this method."

Any workaround?

Many thanks

Soullivaneuh:
Closes #268
Closes #297 
Closes #321

- [x] Make the tests working
- [x] Update the readme according to the package requirements

@lsmith77 This change bump PHP requirement to **7.1** with nelmio usage.

Please tell me what you want for: https://github.com/liip/LiipFunctionalTestBundle/issues/268#issuecomment-340755081

I'll suggest to make it on a new major after waiting changes to clean other dependencies and bump our version support.

If you just think this change acceptable for a major, then merge it when ready and release it. :wink:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,11 +30,10 @@
         "doctrine/doctrine-fixtures-bundle": "~2.3",
         "phpunit/phpunit": "^4.8.35|^5.7|^6.1",
         "twig/twig": "~1.12|~2.0",
-        "nelmio/alice": "~1.7|~2.0",
+        "theofidry/alice-data-fixtures": "^1.0",
         "jackalope/jackalope-doctrine-dbal": "1.1.*|1.2.*",
         "doctrine/phpcr-odm": "~1.3",
         "doctrine/phpcr-bundle": "~1.3",
-        "hautelook/alice-bundle": "~0.2|~1.2",
         "brianium/paratest": "~0.12.0|~0.13.2",
         "doctrine/data-fixtures": "1.2.2"
     },
--- a/README.md
+++ b/README.md
@@ -404,8 +404,9 @@ Tips for Fixture Loading Tests
 If you would like to setup your fixtures with yml files using [Alice](https://github.com/nelmio/alice),
 [`Liip\FunctionalTestBundle\Test\WebTestCase`](Test/WebTestCase.php) has a helper function `loadFixtureFiles`
 which takes an array of resources, or paths to yml files, and returns an array of objects.
-This method uses the [Alice Loader](https://github.com/nelmio/alice/blob/master/src/Nelmio/Alice/Fixtures/Loader.php)
-rather than the FunctionalTestBundle's load methods. You should be aware that there are some difference between the ways these two libraries handle loading.
+This method uses the [Theofidry AliceDataFixtures loader](https://github.com/theofidry/AliceDataFixtures#doctrine-orm)
+rather than the FunctionalTestBundle's load methods.
+You should be aware that there are some difference between the ways these two libraries handle loading.
 
 ```php
 $fixtures = $this->loadFixtureFiles(array(
@@ -445,38 +446,6 @@ $files = array(
  );
 $fixtures = $this->loadFixtureFiles($files, true, null, 'doctrine', ORMPurger::PURGE_MODE_TRUNCATE );
 ```
-
-#### HautelookAliceBundle Faker Providers
-
-This bundle supports faker providers from HautelookAliceBundle.
-Install the bundle with `composer require --dev hautelook/alice-bundle:~1.2` and use the
-[HautelookAliceBundle documentation](https://github.com/hautelook/AliceBundle/blob/1.x/src/Resources/doc/faker-providers.md#faker-providers)
-in order to define your faker providers.
-
-You'll have to add the following line in the `app/AppKernel.php` file:
-
-```php
-<?php
-// app/AppKernel.php
-
-// ...
-class AppKernel extends Kernel
-{
-    public function registerBundles()
-    {
-        // ...
-        if (in_array($this->getEnvironment(), array('dev', 'test'))) {
-            $bundles[] = new Hautelook\AliceBundle\HautelookAliceBundle();
-        }
-
-        return $bundles;
-    }
-
-    // ...
-}
-```
-
-Then you can load fixtures with `$this->loadFixtureFiles(array('@AcmeBundle/…/fixture.yml'));`.
 
 ### Non-SQLite
 
--- a/src/Test/WebTestCase.php
+++ b/src/Test/WebTestCase.php
@@ -530,15 +530,13 @@ abstract class WebTestCase extends BaseWebTestCase
      */
     public function loadFixtureFiles(array $paths = [], $append = false, $omName = null, $registryName = 'doctrine', $purgeMode = null)
     {
-        if (!class_exists('Nelmio\Alice\Fixtures')) {
-            // This class is available during tests, no exception will be thrown.
-            // @codeCoverageIgnoreStart
-            throw new \BadMethodCallException('nelmio/alice should be installed to use this method.');
-            // @codeCoverageIgnoreEnd
-        }
-
         /** @var ContainerInterface $container */
         $container = $this->getContainer();
+
+        $persisterLoaderServiceName = 'fidry_alice_data_fixtures.doctrine.persister_loader';
+        if (!$container->has($persisterLoaderServiceName)) {
+            throw new \BadMethodCallException('theofidry/alice-data-fixtures must be installed to use this method.');
+        }
 
         /** @var ManagerRegistry $registry */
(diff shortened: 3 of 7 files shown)


Test/WebTestCase.php
```diff
@@ -552,16 +552,14 @@ private function locateResources($paths)
...
-            // @codeCoverageIgnoreStart
-            throw new \BadMethodCallException('nelmio/alice should be installed to use this method.');
-            // @codeCoverageIgnoreEnd
-        }
-
         /** @var ContainerInterface $container */
         $container = $this->getContainer();
 
+        $persisterLoaderServiceName = 'fidry_alice_data_fixtures.doctrine.persister_loader';
+        if (!$container->has($persisterLoaderServiceName)) {
```
Soullivaneuh: @theofidry This is always returning false even with the enabled bundle: https://github.com/liip/LiipFunctionalTestBundle/pull/337/files#diff-5658229c21833f6b720a1c07e3705aa0R26

Any idea?

Soullivaneuh: I know why: Wrong kernel. :troll:

Soullivaneuh:
@lsmith77 As you can see the custom faker provider is still tested without hautelook and works like a charm. So we don't need it anymore.

Also, StyleCI is crying about unrelated file: https://styleci.io/analyses/qv5Dmd

Could you please make a fix on master so I can make a rebase? Thanks.

Test/WebTestCase.php
```diff
@@ -574,18 +572,7 @@ public function loadFixtureFiles(array $paths = array(), $append = false, $omNam
 
         $files = $this->locateResources($paths);
 
-        // Check if the Hautelook AliceBundle is registered and if yes, use it instead of Nelmio Alice
-        $hautelookLoaderServiceName = 'hautelook_alice.fixtures.loader';
```
theofidry: I can't recall if that is purged, if it is then the above service should be changed for `fidry_alice_data_fixtures.loader.doctrine`

Tests/AppConfig/config.yml
```diff
@@ -29,4 +29,4 @@ liip_functional_test:
 services:
     faker.provider.foo:
         class: Liip\FunctionalTestBundle\Tests\AppConfig\DataFixtures\Faker\Provider\FooProvider
-        tags: [ { name: hautelook_alice.faker.provider } ]
```
theofidry: the doc needs to be updated accordingly

composer.json
```diff
@@ -52,6 +51,8 @@
             "Liip\\FunctionalTestBundle\\": ""
         }
     },
+    "minimum-stability": "beta",
```
theofidry: rc? (not sure that works)

composer.json
```diff
@@ -30,11 +30,10 @@
         "doctrine/doctrine-fixtures-bundle": "~2.3",
         "phpunit/phpunit": "^4.8.35|^5.7|^6.1",
         "twig/twig": "~1.12|~2.0",
-        "nelmio/alice": "~1.7|~2.0",
+        "theofidry/alice-data-fixtures": "^1.0",
```
theofidry: you also need to dump php to `^7.1` (requirement of `theofidry/alice-data-fixtures`)

Tests/AppConfig/config.yml
Soullivaneuh: See my todos.

composer.json
Soullivaneuh: I don't know, this is why I'm talking about a new major on the thread.

Indeed, PHP 7.1 is required for this dependency, but alice-data-fixtures is not on this bundle.

composer.json
Soullivaneuh: `RC`. I used `beta` before because I was required hautelook.

Test/WebTestCase.php
Soullivaneuh: @theofidry You mean this? https://github.com/liip/LiipFunctionalTestBundle/pull/337/files#diff-f815866dba659cf4e7aebb6fc769c44cR558

Right?

theofidry: yep

Soullivaneuh:
@lsmith77 I think this should be done on a new major with other updates, because it's finally BC break at end usage.

Indeed, the dependency is optional. So, someone making a `composer update` will get the next minor release of this bundle and will see all the tests suites break because of this "dependency" change.

Plus, if the concerned user is not under PHP 7.1, he would not be able to fix it and will have to change the liip dependency to `1.8.*` which is not semver compliant.

What do you think?

Soullivaneuh:
@lsmith77 The PR is finished for the first part. I'm waiting Travis and your thinking about the possible next major release.

Test/WebTestCase.php
Soullivaneuh: Well, there is nothing more to do in this case. :-)

Tests/AppConfig/config.yml
Soullivaneuh: @theofidry I removed the doc part concerning the faker provider. This part was only telling about hautelook installation which is not needed anymore.

Soullivaneuh:
@theofidry Alice data-fixtures looks to not be compatible with SF 2.7: https://travis-ci.org/liip/LiipFunctionalTestBundle/jobs/295358260#L590

Could you confirm?

BTW, it looks like the bridge integration is not tested according to the dependencies: https://packagist.org/packages/nelmio/alice#v3.1.1

theofidry:
>Alice data-fixtures looks to not be compatible with SF 2.7: https://travis-ci.org/liip/LiipFunctionalTestBundle/jobs/295358260#L590

It's not incompatible per se but that compatibility is not guaranteed neither tested for both AliceDataFixtures and NelmioAlice.

>BTW, it looks like the bridge integration is not tested according to the dependencies

There is a dep for local tests but for Travis the deps are overriden, so you can see it in the `.travis.yml`

Soullivaneuh:
> It's not incompatible per se but that compatibility

So we also have to drop SF 2.7 dependency.

theofidry:
Or add it back in the downstream libraries :)

But I'm more in favour of moving forward. We are bumping a bit everything and the old compatible stuff is not going away.

composer.json
```diff
@@ -52,6 +51,8 @@
             "Liip\\FunctionalTestBundle\\": ""
         }
     },
+    "minimum-stability": "RC",
```
lsmith77: why did you add this?

lsmith77:
I also think this Bundle hasn't seen major updates in a long time, so I am not so worried about shedding some old stuff. That being said, we also talked about some other major refactorings to decouple the huge base class a bit.

composer.json
theofidry: theofidry/alice-data-fixtures is still in RC, so if not provided the user would need to install it himself and the require would not suffice

Soullivaneuh:
@lsmith77 In this case I'll prepare some separated PR about version support dropping.

I assume you will create a 1.x branch and let master for 2.x?

lsmith77:
yeah .. a 1.x branch makes sense .. you have the permissions to do this yourself right?

Soullivaneuh:
@lsmith77 I don't think so, I'm not a maintainer of this bundle. :sweat_smile:

Soullivaneuh:
@lsmith77 Please see: https://github.com/liip/LiipFunctionalTestBundle/issues/338

alexislefebvre:
@Soullivaneuh I fixed the StyleCI warnings on the `master` branch. Rebasing your branch on the `master` branch should remove these warnings.

composer.json
Jean85: If we retarget this PR to the 2.x, requiring 7.1 is already included.

Soullivaneuh:
@alexislefebvre I did a rebase, sorry for the delay.

Jean85:
@Soullivaneuh you're still targeting master, that what you want to do? I already stated that I wouldn't do it...

Soullivaneuh:
@Jean85 oh right! It's a mistake. I'm used to have master branch targeting unstable release. I'll rebase it.

composer.json
Soullivaneuh: Rebased.

Jean85:
Ok, great! So we shall close this?

Soullivaneuh:
We can.

PR #339, opened 2017-11-02
Paratest fixes

kick-the-bucket:
* Prevent the application from exiting after schema creation
* Prevent paratest from time-outing


kick-the-bucket:
The ParatestCommandTest is actually broken for all Symfony versions, because without the `$application->setAutoExit(false);` no assertions were ever executed, since the test exited along with the schema creation command.

alexislefebvre:
Can you please open another PR that contain only the StyleCI fixes? It will be easier to review the changes related to paratest, without the unrelated fixes. Thanks.

(if you open another PR, I'll do it in the next days)

kick-the-bucket:
Will do tomorrow.

alexislefebvre:
Was it required to add the `User` entity?

kick-the-bucket:
The command does output errors when it doesn't find any fixtures, but I
guess the tests could pass without them...

kick-the-bucket:
Any ides about the phpunit crashes in the build?

Tests/Command/ParatestCommandTest.php
```diff
@@ -20,15 +20,15 @@
  * Tests/App/AppKernel.php.
  * So it must be loaded in a separate process.
  *
- * @runTestsInSeparateProcesses
+ * @runInSeparateProcess
```
alexislefebvre: Please revert this, this annotation is only for functions: https://phpunit.de/manual/current/en/appendixes.annotations.html#appendixes.annotations.runInSeparateProcess

It may fix the error with the failing test.

alexislefebvre:
@kick-the-bucket We have updated the test matrix and fixed bugs with versions of dependencies so you can rebase your PR and tests should work again.

Tests/Command/ParatestCommandTest.php
kick-the-bucket: Reverted

kick-the-bucket:
Simplified my PR as much as possible, but the same test still fails on PHP 5.6

kick-the-bucket:
Seems like this is the culprit - minimum Twig version is probably too low:
`Fatal error:  Call to undefined function twig_upper_filter()`

alexislefebvre:
I"m closing this since support of paratest has been dropped in #560

PR #346, opened 2017-12-01
allow Symfony 4

lsmith77:


alexislefebvre:
That's strange, it looks like Symfony 3.4.0 is broken on Travis CI.

.travis.yml
```diff
@@ -28,10 +29,14 @@ matrix:
       env: SYMFONY_VERSION="3.2.*"
     - php: 7.1
       env: SYMFONY_VERSION="3.3.*"
+    - php: 7.2
+      env: SYMFONY_VERSION="3.4.*"
+    - php: 7.2
+      env: SYMFONY_VERSION="4.0.*"
```
alexislefebvre: I'm not sure what we'll be able to support Symfony 4.0, maybe we should only support the LTS versions (2.7, 2.8 and 3.4) for the moment.

lsmith77: I will play around with it a bit ..
That being said .. some of the DI based features of this Bundle likely need to change for Symfony 4 to really make sense ..

but if we don’t support Symfony 4 .. then all Bundles using our code here will have a problem.

alexislefebvre:
I can reproduce the error on my computer, here is the content of the file that cause the error:

```
<?php

use Symfony\Component\DependencyInjection\Argument\RewindableGenerator;

// This file has been auto-generated by the Symfony Dependency Injection Component for internal use.
// Returns the public 'console.command_loader' shared service.

return $this->services['console.command_loader'] = new \Symfony\Component\Console\CommandLoader\ContainerCommandLoader(new \Symfony\Component\DependencyInjection\ServiceLocator(array('console.command.about' => function () {
    $f = function (\Symfony\Bundle\FrameworkBundle\Command\AboutCommand $v) { return $v; }; return $f(${($_ = isset($this->services['console.command.about']) ? $this->services['console.command.about'] : $this->load(__DIR__.'/getConsole_Command_AboutService.php')) && false ?: '_'});
}, 
```

It looks like `return $this->services['console.command_loader'] = new …` is causing the error.

alexislefebvre:
When i try to rerun only one test in order to identify the error, then the error disappear!

```bash
./vendor/bin/phpunit --filter ConfigurationConfigTest
PHPUnit 6.5.1 by Sebastian Bergmann and contributors.

...........                                                       11 / 11 (100%)

Time: 1.33 seconds, Memory: 8.00MB

OK (11 tests, 33 assertions)

Remaining deprecation notices (1)
```

alexislefebvre:
The fact that the error only appeared when all the test suite made me think that it was due to the fact that tests are launched in separate processes (due to different Kernel which load different configuration).

I think that I found a solution:

Replace this:
```php
 * @runTestsInSeparateProcesses
```

With this:
```php
 * @runTestsInSeparateProcesses
 * @preserveGlobalState disabled
```

.travis.yml
alexislefebvre: You're right. But maybe we can focus this PR on supporting only LTS versions (so that users can use Symfony 2.7, 2.8 or 3.4 today) and work on the Symfony 4.0 support in another PR.

alexislefebvre:
Using `@preserveGlobalState disabled` solve the PHP errors: https://github.com/alexislefebvre/LiipFunctionalTestBundle/pull/52

But the `testRunCommandVerbosity*` tests are broken with Symfony 3.4. It looks like the verbosity the way used to change verbosity does not work anymore.

weaverryan:
Sorry to complicate things, but there is one issue you guys may not be aware of. In DoctrineFixturesBundle 3.0 (which is what people will need to start using in a bundleless setup), fixture classes are actually tagged services. This creates a problem with this bundle, as it tries to instantiate the fixtures. Instead, for DoctrineFixturesBundle 3.0, in a compiler pass, we call `addFixtures()` on the loader service, passing it all tagged services.

One solution might be to create your own service (in a compiler pass) that is passed all tagged repositories. This class would be a simple map of the repo objects: it could have a method like `getFixture($fixtureClass)` on it. Then, when the user passes the array of fixture classes, you would use that service to fetch off only the fixtures objects you want. Then, create the loader:

```php
$loader = new SymfonyFixturesLoader($container);
$loader->addFixtures($fixtureObjects);
```

Hopefully the other fixtures libraries will follow this new pattern, and eventually it might simplify some logic in the library. If you have any questions, please let me know!

lsmith77:
I am beginning more and more to accept that maybe its not feasible to bring this Bundle to v4 .. and maybe we should focus those efforts on making version 2 of this Bundle work with Symfony v3.4+

weaverryan:
Indeed. As nice as this library is, it's loaded with a lot of bloat (well, at least the one giant class is very big). Of course, we need someone to make this effort :)

lsmith77:
https://github.com/liip/LiipFunctionalTestBundle/issues?q=is%3Aopen+is%3Aissue+milestone%3A2.0 :)

alexislefebvre:
@lsmith77 Can I please take your work and work on a new PR focused on the Symfony 3.4 support?

lsmith77:
sure!

lsmith77:
or alternatively we can close this PR in favor of completing #347

for Symfony 4 we focus on 2.0

weaverryan:
@lsmith77 What's the advantage of #347? Dropping 2.7 is a good idea, but it doesn't really help any end users, right? I mean, it seems like we should merge that... but yea... then we're still in a "let's do work fro 2.0 support sf 4.0".

lsmith77:
#347 drops 2.7 but should cover adding support for 3.4

alexislefebvre:
I only want to add 3.4 to the test matrix for the moment, by using code from https://github.com/liip/LiipFunctionalTestBundle/pull/347 and https://github.com/alexislefebvre/LiipFunctionalTestBundle/pull/52. Once this Bundle will work with all the current LTS releases, we'll be able to think about the Symfony 4.0 support.

weaverryan:
👍 let's do it! But not lose track of this!

alexislefebvre:
Oops, I merged #351 that was based on this PR and it closed this PR too.

lsmith77:
no worries.

lsmith77:
for the dependency issues I am following up here https://github.com/liip/LiipFunctionalTestBundle/pull/352

PR #351, opened 2017-12-07
Travis: Support Symfony 3.4 and test PHP 7.2

alexislefebvre:
With commits from #346.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
         "doctrine/common": "~2.0"
     },
     "require-dev": {
-        "symfony/symfony": "^2.7|^3.3|^4.0",
+        "symfony/symfony": "^2.7.1|^3.3|^4.0",
         "symfony/phpunit-bridge": "^2.7|^3.0|^4.0",
         "doctrine/orm": "~2.5",
         "symfony/console": "^2.7|^3.0|^4.0",
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
         "doctrine/common": "~2.0"
     },
     "require-dev": {
-        "symfony/symfony": "^2.7|^3.0|^4.0",
+        "symfony/symfony": "^2.7|^3.3|^4.0",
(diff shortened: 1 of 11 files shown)


alexislefebvre:
I think that I found why verbosity does not work on Symfony 3.4: https://github.com/symfony/symfony/pull/24425

composer.json
```diff
@@ -21,7 +21,7 @@
         "doctrine/common": "~2.0"
     },
     "require-dev": {
-        "symfony/symfony": "^2.7|^3.3|^4.0",
+        "symfony/symfony": "^2.7.1|^3.3|^4.0",
```
alexislefebvre: The `^2.7.1` part is valid for Symfony 2.7 and 2.8: https://semver.mwl.be/#?package=symfony%2Fsymfony&version=%5E2.7.1%7C%5E3.3%7C%5E4.0&minimum-stability=stable

alexislefebvre:
@lsmith77 It looks good to me, except the use of `putenv('SHELL_VERBOSITY='.$level);}` which looks like a strange workaround.

composer.json
lsmith77: 2.7.0 doesn’t work?

lsmith77:
looks good ..
to further improve Symfony 4 support, we could allow phpcr-bundle/phpcr-odm 2.0@dev
but this can come in another PR

composer.json
alexislefebvre: Unfortunately, Symfony 2.7.0 was causing errors in the `--prefer-lowest` job because of deprecations: https://travis-ci.org/liip/LiipFunctionalTestBundle/builds/313161686

alexislefebvre:
And now I realize that the `composer.json` allow Symfony 3.4 and 4 but this version is not compatible with Symfony 4, maybe I should have removed Symfony 4 from `composer.json` …

Anyway we'll have to wait before releasing a new version, otherwise Symfony 4 users will install a version that doesn't work.

PR #347, opened 2017-12-03
Drop support of Symfony < 2.7

alexislefebvre:
Remove outdated code

PR #354, opened 2017-12-13
Changed $_SERVER to $_ENV in Test/WebTestCase.php

AshleyDawson:
When configuring PHPUnit it should be possible to pass `KERNEL_DIR` as an environment variable

### DIFF
--- a/Test/WebTestCase.php
+++ b/Test/WebTestCase.php
@@ -74,7 +74,7 @@ abstract class WebTestCase extends BaseWebTestCase
 
     protected static function getKernelClass()
     {
-        $dir = isset($_SERVER['KERNEL_DIR']) ? $_SERVER['KERNEL_DIR'] : static::getPhpUnitXmlDir();
+        $dir = isset($_ENV['KERNEL_DIR']) ? $_ENV['KERNEL_DIR'] : static::getPhpUnitXmlDir();
 
         list($appname) = explode('\\', get_called_class());
 


alexislefebvre:
Thanks. I don't understand why it doesn't break tests. The variable `$_SERVER['KERNEL_DIR']` is defined in `phpunit.xml.dist`: https://github.com/liip/LiipFunctionalTestBundle/blob/0665c24f9e9b4a7fe018f607eb85f23c0fd89410/phpunit.xml.dist#L22

After your changes this value is ignored, but tests still pass. That's surprising.

alexislefebvre:
Anyway, I'm wondering if we should keep Backward Compatibility.

We can have something like that:

- use `$_SERVER['KERNEL_DIR']` if this value is defined
- use `$_ENV['KERNEL_DIR']` if this value is defined
- else: call `static::getPhpUnitXmlDir()`

alexislefebvre:
You can rebase your branch on `master` to avoid the StyleCI error (I fixed the errors in #347).

AshleyDawson:
Ah, sorry - my mistake, I didn't know you could do `<server name="KERNEL_DIR" value="Tests/App" />`. I should really read the docs. Please reject this PR if necessary.

alexislefebvre:
This PR showed that this feature is documented, thanks for pointing it. We can add a note about this in the README.

BPScott:
I don't think this should be changed, instead it should accept both. This is what the parent's class does in Symfony 3.4 (though using KERNEL_DIR is deprecated): https://github.com/symfony/symfony/blob/77104a1468cb8eb88eb3b699a42bc1dee2be9c46/src/Symfony/Bundle/FrameworkBundle/Test/KernelTestCase.php#L115

That said, Symfony 4.0 removes the concept of the KERNEL_DIR parameter entirely (see https://github.com/symfony/symfony/pull/22675 for the original PR), and says that you should use KERNEL_CLASS instead to explicitly state an autoloadable class name.

With that in mind I don't currently understand why this bundle's WebTestCase needs to overwrite its ancestor's definition of getKernelClass() at all, and I think that whole function definition can be removed entirely.

Issue #353, opened 2017-12-08
Summary of Symfony 4.0 support


alexislefebvre:
It's only a list of issues and PRs related to Symfony 4.0:

- [x] Travis: Support Symfony 3.4 and test PHP 7.2 #351
- [x] Symfony 3.4/4.0 support : KERNEL_DIR is replaced by KERNEL_CLASS #343 
- [x] improve Symfony 4 compatibility #352
- [x] new fixtures loading system: https://github.com/liip/LiipFunctionalTestBundle/pull/346#issuecomment-349081379

Feel free to discuss about problems and solution about the Symfony 4.0 support.

lsmith77:
one of the common issues when doing functional tests with Symfony 4 is that services are now private by default. it might make sense to give some easy way to make a kernel instance forcing specific services to be public.

lsmith77:
we should probably also drop the html5 validation https://github.com/liip/LiipFunctionalTestBundle/issues/251

Jean85:
To access private services, the suggested way is to add a public alias to them, so you don't alter the DI definition with the test, just add a side-access to the instances.

In my tests, I define a `getService($id)` method that adds a `.public_alias` suffix to all my calls, and pre-declare them for all the container; since this bundle caches the container, it's not so heavy to do.

lsmith77:
Yeah that “alias” hack however means we either need to use a different name as you point out .. or we have to rename the original service. imho both are not ideal.

alexislefebvre:
> we should probably also drop the html5 validation #251

I agree.

I try to keep track of the issues related to the next major release in the 2.0 milestone: https://github.com/liip/LiipFunctionalTestBundle/milestone/1

Jean85:
> Yeah that “alias” hack however means we either need to use a different name as you point out .. or we have to rename the original service. imho both are not ideal.

IMHO using my hack is the lesser evil, since it will not alter the app DI configuration. Tests shouldn't alter the SUT.

lsmith77:
I agree in so much as adding an alias is less invasive, however it is still a change to the system. As such the key aspect I see is that such changes can be done on the basis of a single test case, rather than hardcoding it into the functional test kernel for all tests.

Jean85:
Something like a whitelist of services to be rendered public through the public alias?
This would invalidate the whole kernel caching mechanism.

lsmith77:
[...]

PR #355, opened 2017-12-14
Try to load class as a service from the container first.

MonsieurBon:
The PR fixes a compatibility issue with doctrine-fixtures-bundle. Since version 3.0.2, if I need to use services in my fixture classes, I have to define the fixtures as services and use DI to get the service in the constructor. This causes problems with the "new" call in WebTestCase. The PR tries to load the fixtures class as a service from the container first. This fixes the issue for me.

### DIFF
--- a/Test/WebTestCase.php
+++ b/Test/WebTestCase.php
@@ -694,7 +694,13 @@ abstract class WebTestCase extends BaseWebTestCase
      */
     protected function loadFixtureClass($loader, $className)
     {
-        $fixture = new $className();
+        $fixture = null;
+
+        if ($this->getContainer()->has($className)) {
+            $fixture = $this->getContainer()->get($className);
+        } else {
+            $fixture = new $className();
+        }
 
         if ($loader->hasFixture($fixture)) {
             unset($fixture);


Bukashk0zzz:
@MonsieurBon Thanks for PR. Any ideas how to load not public fixtures/services ?

MonsieurBon:
@Bukashk0zzz I don't think it's possible. As far as I understood private/public services, private services are only available through DI and not through the container.

Bukashk0zzz:
@MonsieurBon I google this and also not find any solution. Any way I now use this PR in my code. It  helped a lot.

@alexislefebvre @lsmith77 Can you please review this and merge?

lsmith77:
@Bukashk0zzz I have an idea how we can solve the topic of getting access to private services but this will only happen in 2.0, see #353.

@alexislefebvre this quick fix looks like a good idea to me.

alexislefebvre:
I don't remember how this part of code works so I'm following your advices. :wink:

@MonsieurBon Thanks! Could you please provide an example of how you define fixtures as services? Feel free to submit another PR in order to document this usage of fixtures. It looks like this case is not explained yet: https://github.com/liip/LiipFunctionalTestBundle#database-tests

PR #357, opened 2017-12-22
Register commands manually to avoid SF 3.4 deprecation

Jean85:
This manually registers the commands with the bundle to avoid deprecations in Symfony 3.4.

### DIFF
--- a/DependencyInjection/LiipFunctionalTestExtension.php
+++ b/DependencyInjection/LiipFunctionalTestExtension.php
@@ -11,6 +11,9 @@
 
 namespace Liip\FunctionalTestBundle\DependencyInjection;
 
+use Liip\FunctionalTestBundle\Command\RunParatestCommand;
+use Liip\FunctionalTestBundle\Command\TestCommand;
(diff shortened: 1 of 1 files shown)


Jean85:
This is no longer WIP. Should I bother about StyleCI?

lsmith77:
why did you prefer to register them via PHP rather than an XML file?
as for StyleCI .. not sure why its complaing but generally yeah, it should be green.

lsmith77:
it just passed fine on master with the last merge https://styleci.io/analyses/8P7ndP .. odd

lsmith77:
hmm

Jean85:
I used the PHP approach because I wasn't sure that the XML was included every time...

For the CS, maybe we should add PHP-CS-Fixer to help on the contributor side?

alexislefebvre:
@lsmith77 Do you prefer the commands to be registered through XML?

lsmith77:
yeah. its what is from what I can see the standard.

Jean85:
Closing in favor of #364 with XML.

PR #358, opened 2017-12-22
Simplify Travis matrix

Jean85:
This PR:

 - disallows failures with PHP 7.2: it's now stable and released
 - stops testing for Symfony 3.3, since it's not an LTS and 3.4 is stable too
 - tests for Symfony 3.4 only once
 - stops testing for Symfony 4: we're not ready yet!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,8 +27,6 @@ matrix:
       env: SYMFONY_VERSION="2.8.*"
     - php: 7.2
       env: SYMFONY_VERSION="3.4.*"
-    - php: 7.2
-      env: SYMFONY_VERSION="4.0.*"
   fast_finish: true
 
 before_install:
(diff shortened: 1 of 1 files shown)


alexislefebvre:
Some explanations about your remarks:

We allowed failures from Symfony 4.0, not for PHP 7.2.

Symfony 3.3 is not a LTS but IIRC it is still supported, what's why it was still here.

Symfony 3.4 was tested twice, but with different PHP versions.

Thanks for the clean up.

lsmith77:
ah right .. then again 3.3 is only getting fixes until January https://symfony.com/roadmap?version=3.3#checker

Jean85:
Thanks for the clarifications @alexislefebvre! Btw If it works under 3.0 and 3.4, testing for the intermediate 3.3 becomes not necessary, that's why I removed it.

PR #359, opened 2017-12-22
Prepare release 2.0

Jean85:
This PR attempts to start working toward a 2.0, following the roadmap of #338.

@lsmith77 can you create a dedicated branch for 2.x-dev? So I (and others) can target this PR (and following others) to that branch, without polluting master before it gets somewhat stable.

In here I:

 * move to a more PSR-4 friendly dir structure
 * bumped to PHP 7.1
 * bumped deps to allow only deps compatible with Symfony 4
 * dumped Paratest and Assetic because they aren't

This also includes #357

PR #362, opened 2017-12-27
Use PSR 4 directory structure

alexislefebvre:
Use commit from #359

Replaces #324


alexislefebvre:
@Jean85 @lsmith77 Can you please give your advice about this PR?

I thought that it was better to perform changes on master instead of 2.0 as long as it was possible. And do only BC changes on the 2.0 branch.

lsmith77:
see my response in the other PR, imho its not worth it to try and get those changes into master where we have to ensure 100% BC.

alexislefebvre:
Thanks for your answer, I think that there is no risk of BC break since I only moved some files. The namespace is the same as before.

alexislefebvre:
Well, you're right. We don't need this on master.

PR #363, opened 2017-12-28
Fix Typo

aminebenhariz:
When loading fixture files, setting param "$append" to "true" will keep the old records. Using "false" will clean the tables.

### DIFF
--- a/README.md
+++ b/README.md
@@ -419,7 +419,7 @@ If you want to clear tables you have the following two ways:
 1. Only to remove records of tables;
 2. Truncate tables.
 
-The first way is consisted in using the second parameter `$append` with value `true`. It allows you **only** to remove all records of table. Values of auto increment won't be reset. 
+The first way is consisted in using the second parameter `$append` with value `false`. It allows you **only** to remove all records of table. Values of auto increment won't be reset. 
 ```php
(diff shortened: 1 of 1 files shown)


alexislefebvre:
It looks like it doesn't match with the codes below the changed lines.

alexislefebvre:
Thanks for the second fix.

Here is the documentation after your correction:

> The first way is consisted in using the second parameter `$append` with value `false`.

This is the default value of this parameter, so the explanation is not necessary anymore. I'm wondering how can we rewrite this part.

maximgubar:
> I'm wondering how can we rewrite this part.

@alexislefebvre, what exactly you want to rewrite ?

alexislefebvre:
@maximgubar I checked this PR again and if fixes a mistake but after the changes, the text describes the default behaviour, so I don't think that we should keep it as is. But we may rewrite is as something like “by default, calling `loadFixtureFiles(…)` (without the second parameter) will purge database … calling `loadFixtureFiles(…, true)` will add the data to the existing data”.

alexislefebvre:
I think that I'm overthinking this part of the documentation. This PR fixes an error and it can be merged.

alexislefebvre:
Thanks @aminebenhariz!

PR #365, opened 2018-01-02
[2.x] PSR-4 friendly structure

Jean85:
This PR splits out the structural part from #359. I'm expecting a green build here, since I've only moved files around. Afterward, I'll be opening a second PR with the rest of the work (mainly `composer.json` bumps and build fixes).


alexislefebvre:
Thanks a lot @Jean85 for taking the time to split the previous PR.

Issue #360, opened 2017-12-27
Document usage of KERNEL_DIR/KERNEL_CLASS


alexislefebvre:
See #354

alexislefebvre:
See #343.

PR #366, opened 2018-01-02
[WIP] [2.x] Version bumps

Jean85:
This PR also stems from #359; it includes #365, so I will rebase it as soon as it gets merged.

Currently, I'm unsure on how to proceed to resolve the `KERNEL_CLASS` issue, since the `KERNEL_DIR` variabile will not work under SF 4. I would like to drop completely `WebTestCase` and fall back to the Symfony native behavior, but it may not work for the multiple kernels that we have in the tests. Also, #360 seems to require it... @alexislefebvre what do you think?


Jean85:
Rebased onto the new PSR-4 structure after that #365 got merged.

alexislefebvre:
I don't understand the _LogicException: Cannot register "Hautelook\AliceBundle\HautelookAliceBundle" without "Fidry\AliceDataFixtures\Bridge\Symfony\FidryAliceDataFixturesBundle"._ error. It looks like it was addressed in #321 and #337.

Jean85:
Well, both are not merged! I'll try to start from there then.

composer.json
```diff
@@ -59,7 +59,7 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "2.x-dev"
+            "dev-master": "1.7.x-dev"
```
alexislefebvre: I think that this should not be changed.

Jean85: I'm simply reverting a change that predates the creation of the `2.x` branch, which is a Composer-friendly name. `master` is stil on the 1.x for now, until we merge the `2.x` onto it.

alexislefebvre: That's a interesting point, I don't know if we'll merge 2.x once the work on this branch will be over.

Anyway, I'm still confused about the `1.7.x-dev` since we are on the 2.x branch.

And even if we use `1.x` as a branch alias, shouldn't it be `1.9.x`? The last version is `1.9.2`: https://packagist.org/packages/liip/functional-test-bundle

Soullivaneuh:
No BC layer are removed on from the code here. I'm working on it on a new branch.

Jean85:
Closing in favor of #379

Issue #270, opened 2016-06-09
You have requested a non-existent service "liip_functional_test.validator".


Chrysweel:
Hello.

I just installed this bundle, with instructions of the documentation.

And with test of example I get the following error:

```
Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException: You have requested a non-existent service "liip_functional_test.validator".

/mypath/app/bootstrap.php.cache:2063
/mypath/vendor/liip/functional-test-bundle/Utils/HttpAssertions.php:90
/mypath/vendor/liip/functional-test-bundle/Test/WebTestCase.php:882
```

In my composer.json I have:

```
"symfony/symfony" : "2.3.*",
"liip/functional-test-bundle" : "^1.6"
```

AppKernel.

```
        if (in_array($this->getEnvironment(), array('dev', 'test'))) {
            // $bundles[] = new Symfony\Bundle\WebProfilerBundle\WebProfilerBundle();
            $bundles[] = new Sensio\Bundle\DistributionBundle\SensioDistributionBundle();
            $bundles[] = new Sensio\Bundle\GeneratorBundle\SensioGeneratorBundle();
            $bundles[] = new Liip\FunctionalTestBundle\LiipFunctionalTestBundle();
        }
```

Thanks in advance !

alexislefebvre:
Can you please add the code from your test?

Chrysweel:
Yes.

```
<?php

namespace MyProject\CoreBundle\Tests\Controller;

use Liip\FunctionalTestBundle\Test\WebTestCase;

class LipControllerTest extends WebTestCase
{
    public function testContact()
    {
        $client = static::makeClient();
        $crawler = $client->request('GET', '/contacto');
        $this->assertStatusCode(200, $client);
        $form = $crawler->selectButton('Enviar')->form();
        $crawler = $client->submit($form);

        // We should get a validation error for the empty fields.
        $this->assertStatusCode(200, $client);
        $this->assertValidationErrors(['data.email', 'data.message'], $client->getContainer());

        // Try again with with the fields filled out.
        $form = $crawler->selectButton('Submit')->form();
        $form->setValues([
        		'contact[name]'  	=> 'Hello',
        		'contact[email]' 	=> 'nobody@example.com',
        		'contact[subject]'	=> 'asunto',
        		'contact[body]'		=> 'body'
        		
        ]);
        $client->submit($form);
        $this->assertStatusCode(302, $client);
    }
}
```

[...]

PR #367, opened 2018-01-09
#270 fix non-existent service in SF3.4/4.0

cacahouwete:
#270 declare services as public ! By default services are private since sf3.4
https://symfony.com/blog/new-in-symfony-3-4-services-are-private-by-default


alexislefebvre:
Thanks!

PR #370, opened 2018-01-10
FlintCI integration

Soullivaneuh:
As proposed on https://github.com/liip/LiipFunctionalTestBundle/issues/338#issuecomment-356663407, here is a FlintCI configuration: 

I also took permissions to add some extra CS rules on .php_cs. Tell me if you want to change some!

To enable FlintCI, just connect here: https://flintci.io and enable your repository. You may want to change some settings. After done, the first analysis will be done on next push. Closing and re-opening this PR may do the thing too.

As you can see, this way permit to access the full configuration of php-cs-fixer v2 and both cli and diff view. Plus, other services are available like yamllint.

The interface is still quite primitive, I'm on heavy development of the app. Don't hesitate to give some feedbacks! :+1: https://gitlab.com/flintci/flintci

Note: I let the StyleCI configuration for now. I let you decide to keep both or completely migrate to FlintCI.

Regards.

PR #373, opened 2018-01-10
[WIP] Remove outdated code in Html5WebTestCase

### LINKED ISSUE #251: Remove old code with comment "this stuff should go away"?

alexislefebvre:
Fixes #251 

Maybe we should remove all the code related to HTML5 validation? What do you think @lsmith77 and others? Thanks.


lsmith77:
yes. lets kick the html5 validation feature. if at all it should live in its own bundle.

alexislefebvre:
Great, I'll remove all the HTML validation part. I'll have to document that somewhere.

Jean85:
You also need to include #370 to avoid conflicts.

PR #378, opened 2018-01-13
Remove kernel dir trick

Soullivaneuh:
First, because it's deprecated and removed in Symfony 4.0 as the kernel is now auto-loaded like any another PHP class.

Second, because I don't understand why we need this at all, and I'm not the only one: https://github.com/liip/LiipFunctionalTestBundle/pull/354#issuecomment-352617286

Plus, we can't find anything in the documentation about that part. (#360)

So expect you can give a real reason of why we need this, we should remove it on next major. It's the only blocking thing to get this running under SF4.

### DIFF
--- a/src/Test/WebTestCase.php
+++ b/src/Test/WebTestCase.php
@@ -50,8 +50,6 @@ abstract class WebTestCase extends BaseWebTestCase
 
     protected $containers;
 
-    protected $kernelDir;
-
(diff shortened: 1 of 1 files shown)


Soullivaneuh:
Integrated in #379

Jean85:
Should we close this too?

Soullivaneuh:
Yes.

PR #174, opened 2015-10-21
Append fixtures

pstephenson02:
While I don't necessarily want to add more to this loadFixtures method, this functionality is important to the test suite I work on. I can add documentation pending any changes/updates during review.

PR #382, opened 2018-01-18
[2.x] Upgrade guide

Jean85:
This starts to add info to the changelog. I watched over the diff between the branches to determine it, but please review it.

I started the upgrade guide, but I'm not sure how to structure it... Suggestions?

PR #384, opened 2018-01-21
add append fixtures feature to loadFixtures method

almacbe:
Add an optional parameter that allows to append fixtures.

As we talked on #174 I have implemented a proposal for that feature.

### DIFF
--- a/UPGRADE-2.0.md
+++ b/UPGRADE-2.0.md
@@ -8,3 +8,7 @@ This is the list of actions that you need to take when upgrading this bundle fro
 composer remove --dev nelmio/alice
 composer require --dev liip/functional-test-bundle "~2.0"
 ```
+
+ * The interface of `LoadFixtures` had to be changed to allow append fixtures. The main difference is it had been added
+ a boolean second parameter. You will have to add it to `false` if you had changed the default manager, driver
+ or purge mode.
--- a/src/Test/WebTestCase.php
+++ b/src/Test/WebTestCase.php
@@ -309,13 +309,14 @@ abstract class WebTestCase extends BaseWebTestCase
      * class path.
      *
      * @param array  $classNames   List of fully qualified class names of fixtures to load
+     * @param bool   $append
      * @param string $omName       The name of object manager to use
      * @param string $registryName The service id of manager registry to use
      * @param int    $purgeMode    Sets the ORM purge mode
      *
      * @return null|AbstractExecutor
      */
-    protected function loadFixtures(array $classNames = [], ?string $omName = null, string $registryName = 'doctrine', ?int $purgeMode = null): ?AbstractExecutor
+    protected function loadFixtures(array $classNames = [], bool $append = false, ?string $omName = null, string $registryName = 'doctrine', ?int $purgeMode = null): ?AbstractExecutor
     {
         $container = $this->getContainer();
         /** @var ManagerRegistry $registry */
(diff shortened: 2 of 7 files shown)


alexislefebvre:
Adding the option as a second parameter will break many calls to `loadFixtures()`. But it's consistent with the order of parameters in https://github.com/liip/LiipFunctionalTestBundle/blob/48ec7c61ebcaf04955926a985b67d0389137a0ae/src/Test/WebTestCase.php#L508 So I agree with this change. :+1:

almacbe:
That was my thought

Jean85:
Since this change is breaking BC, we should add this information in the upgrade guide and the changelog (see #382)

almacbe:
Then, I am sorry but I don't know what I have to do... Do I have to wait? or do I have to modify the files?

Jean85:
It's up to @alexislefebvre or @lsmith77 to decide if they want to merge this and do it later, or merge #382 and do it now in this same PR..

Jean85:
#382 is merged, you can pull from the target branch and add info in those files now! :+1:

almacbe:
I have updated the doc, changelog and update guide. I hope my English it is good enough.

almacbe:
@Jean85 Do I have to change or do something more?

UPGRADE-2.0.md
```diff
@@ -8,3 +8,7 @@ This is the list of actions that you need to take when upgrading this bundle fro
 composer remove --dev nelmio/alice
 composer require --dev liip/functional-test-bundle "~2.0"
 ```
+
+ * It had been changed the interface of `LoadFixtures` to allow append fixtures. The main difference is it had been added
```
lsmith77: just a small wording tweak

`The interface of `LoadFixtures` had to be changed to allow append fixtures.`

src/Test/WebTestCase.php
```diff
@@ -309,13 +309,14 @@ protected function isBackupUpToDate(array $classNames, string $backup): bool
...
      * @param array  $classNames   List of fully qualified class names of fixtures to load
+     * @param bool   $append
      * @param string $omName       The name of object manager to use
      * @param string $registryName The service id of manager registry to use
      * @param int    $purgeMode    Sets the ORM purge mode
      *
      * @return null|AbstractExecutor
      */
-    protected function loadFixtures(array $classNames = [], ?string $omName = null, string $registryName = 'doctrine', ?int $purgeMode = null): ?AbstractExecutor
+    protected function loadFixtures(array $classNames = [], bool $append = false, ?string $omName = null, string $registryName = 'doctrine', ?int $purgeMode = null): ?AbstractExecutor
```
lsmith77: generally this method has too many parameters already from my point of view. it would be good to think about how to refactor it to have less parameters.

UPGRADE-2.0.md
almacbe: @lsmith77 Thanks, it is sorted.

src/Test/WebTestCase.php
Jean85: Since we are targeting 2.x and BC are allowed, it's a good moment to do that, yes! :+1:

almacbe: But, has it to be in this PR or another one?

almacbe:
Do I have to do or change something? @lsmith77 @Jean85

lsmith77:
@alexislefebvre if this approach is ok for you, then I can live with it .. we should just immediate start this follow up PR or at least create a ticket and someone should feel somewhat responsible for taking it on.

alexislefebvre:
I prefer that we refactor `loadFixtures()` and `loadFixtureFiles()` in another PR too.

PR #388, opened 2018-02-02
fixes #387

kalyabin:
*Php version:* 7.1.6
*Phpunit version:* 7.0.0
*LiipFunctionalTestBundle version*: 1.9.2

After update phpunit from 5.5.4 to 7.0.0 my test case ends with fatal error:

*TestCase*:

```php
class RememberPasswordControllerTest extends WebTestCase
{
    ...
    public function testSomeRequest()
    {
        $client = $this->createClient();

        $url = $this->getUrl('remember_password.change_password', [
            'checkerId' => '...',
            'code' => '...',
        ]);

        $client->request('POST', $url);

       $this->assertStatusCode(400, $client);
    }
    ...
}
```

There was an error:

```
TypeError: Argument 3 passed to PHPUnit\Framework\Assert::assertEquals() must be of the type string, null given, called in /var/www/.../vendor/liip/functional-test-bundle/Utils/HttpAssertions.php on line 81

/var/www/.../vendor/liip/functional-test-bundle/Utils/HttpAssertions.php:81
/var/www/.../vendor/liip/functional-test-bundle/Test/WebTestCase.php:928
/var/www/.../tests/UserBundle/Tests/Controller/RememberPasswordControllerTest.php:230
```

\PHPUnit\Framework\Assert uses strict type for third argument at [Assert::assertEquals](https://github.com/sebastianbergmann/phpunit/blob/master/src/Framework/Assert.php#L576) method.

PR #390, opened 2018-02-02
Declare conflict with PHPUnit 7

Jean85:
As discovered in https://github.com/liip/LiipFunctionalTestBundle/pull/388#issuecomment-362587838, we are not compatible with PHPUnit 7, we should declare it in the `composer.json` for the 1.x releases.

The annoying stuff is that the old releases will be missing this info forever. Should we do something else? Add a "known issue" somewhere?

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,9 @@ Installation
     This command requires you to have Composer installed globally, as explained
     in the [installation chapter](https://getcomposer.org/doc/00-intro.md)
     of the Composer documentation.
+    
+    Note: if you are using PHPUnit 7 or later, you MUST require at least version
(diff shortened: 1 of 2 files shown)


alexislefebvre:
Maybe we should have added something like `phpunit/phpunit:~5|~6` in the `require` part of `composer.json` in order to restrict the PHPUnit versions in previous releases. But as you noted, it's too late for the old releases, we can't change them.

---

> Add a "known issue" somewhere?

I think that users will understand quickly that updating PHPUnit to 7 causes the error mentioned in the comment. But mentioning it in the documentation is a good idea.

alexislefebvre:
That's perfect, thanks!

PR #393, opened 2018-02-08
add append feature for sqlite on load fixtures method

almacbe:
I realize that I forget to add the append feature to SQLite. I've added it to this PR.

### DIFF
--- a/src/Test/WebTestCase.php
+++ b/src/Test/WebTestCase.php
@@ -383,9 +383,11 @@ abstract class WebTestCase extends BaseWebTestCase
 
                 // TODO: handle case when using persistent connections. Fail loudly?
                 $schemaTool = new SchemaTool($om);
-                $schemaTool->dropDatabase();
-                if (!empty($metadatas)) {
-                    $schemaTool->createSchema($metadatas);
+                if (false === $append) {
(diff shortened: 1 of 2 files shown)


almacbe:
And... how could I test this feature? I am not sure how to create a test in an environment with SQLite config.

alexislefebvre:
Default configuration of tests uses SQLite:

https://github.com/liip/LiipFunctionalTestBundle/blob/ba8e654c798619aea58de1d3b912bcb718135340/tests/App/config.yml#L35

There is also some tests with MySQL, they are marked with `@group mysql` and they are ignored by default:

https://github.com/liip/LiipFunctionalTestBundle/blob/3bca1f2349c694f16670d6de63020c28c55df9fa/tests/Test/WebTestCaseConfigMysqlTest.php#L61

But Travis CI launch them:

https://github.com/liip/LiipFunctionalTestBundle/blob/48ec7c61ebcaf04955926a985b67d0389137a0ae/.travis.yml#L46

So as long as a test use the append feature, it will be tested with SQLite.

almacbe:
I have added the test.

almacbe:
Do you have some update about this PR?

alexislefebvre:
Thanks!

Issue #392, opened 2018-02-08
[2.0] Reduce size of WebTestCase?


alexislefebvre:
# Problem

As [suggested](https://github.com/liip/LiipFunctionalTestBundle/issues/381#issuecomment-358399299) by @Soullivaneuh, we can discuss here about the size problem of `WebTestCase`.

The last version of `WebTestCase` contains 888 lines, including 28 `use` declarations: https://github.com/liip/LiipFunctionalTestBundle/blob/3bca1f2349c694f16670d6de63020c28c55df9fa/src/Test/WebTestCase.php

**This class has accumulated features since 7 years, this is nice but it looks more and more like a [Big Ball Of Mud](http://wiki.c2.com/?BigBallOfMud). We have the opportunity to clean it up.**

---

I identified some groups of related methods (methods marked with ¹ were moved in the `HttpAssertions` one year ago):

- general utilities: `getKernelClass()`, `getServiceMockBuilder()`, `getContainer()`
- loading fixtures: `loadFixtures()`, `loadFixtureFiles()`, `setExcludedDoctrineTables()` and all the private methods that are needed by these methods
- running commands: `runCommand()`, `setVerbosityLevel()`
- authentication:  `createUserToken()`, `loginAs()`
- request helpers and HTTP assertions: `makeClient()`, `isSuccessful()`¹, `fetchContent()`, `fetchCrawler()`, `assertStatusCode()`¹, `assertValidationErrors()`¹

We're very far from [Single responsibility principle](https://en.wikipedia.org/wiki/Single_responsibility_principle). 😅

Interestingly, about one half of the class code is used to load fixtures. So I think that it should be the first thing to move.

# Solution

@lsmith77 has suggested several possibilities:
- traits: https://github.com/liip/LiipFunctionalTestBundle/issues/332#issuecomment-329958638
- services: refactor Bundle to provide all features as services #6 (this one is really old 😃)

## Traits

I was in favour of traits but @Jean85 added [a strong argument againt it](https://github.com/liip/LiipFunctionalTestBundle/issues/381#issuecomment-358574095):

> I'm personally against traits. They are simple a dirty way to avoid copy and paste, and I don't think we have many places were we should use it.

With traits, complexity would persist since `WebTestCase` and its traits would still be linked together.

## Services

[...]

PR #398, opened 2018-02-14
Separate db tool and db backup/restore services for version 2.x

alekseytupichenkov:
Provide services for backup/restore different database and refactoring code


alekseytupichenkov:
Also, I have suggestion, maybe need remove `liip_functional_test.cache_sqlite_db` at all? Currently we have `liip_functional_test.cache_db.*` to make this

alexislefebvre:
There is a conflict since #393 has been merged, sorry.

Thanks for the amazing work, it looks really nice!

I think that we can drop `liip_functional_test.cache_sqlite_db` if you added a replacement, we don't need to keep a BC parameter on the 2.x branch.

src/Services/DatabaseBackup/MysqlCustomDatabaseBackup.php
```diff
@@ -0,0 +1,73 @@
...
+
+        $dbName = isset($params['dbname']) ? $params['dbname'] : '';
+        $dbHost = isset($params['host']) ? $params['host'] : '';
+        $dbPort = isset($params['port']) ? $params['port'] : '';
+        $dbUser = isset($params['user']) ? $params['user'] : '';
+        $dbPass = isset($params['password']) ? $params['password'] : '';
+
+        $executor->getReferenceRepository()->save($this->getBackupName());
+
+        exec("mysqldump -h $dbHost -u $dbUser -p$dbPass $dbName > {$this->getBackupName()}");
```
alexislefebvre: If the MySQL database is in a Docker container, this command won't work?

powerpd: I use docker too, it will works fine if you install mysql-client in container with symfony app :)

powerpd:
I'll resolve conflicts in this weekend

alekseytupichenkov:
@alexislefebvre done :)
Also updated pr for 1.9.x

alexislefebvre:
Can you please explain why you call `mysqldump`? I don't understand why you chose to rely on a CLI command, the current version of bundle only rely on PHP code and I prefer this option personnally.

Otherwise, this PR is really good.

@Jean85 @lsmith77 Could you please give your opinion about this PR ? Thanks.

alexislefebvre:
I think that the backup is done with the `save()` function: https://github.com/liip/LiipFunctionalTestBundle/blob/afddaec6953fa0d2217c1aa67a49da5259503a1e/src/Test/WebTestCase.php#L435

Here is the method: https://github.com/doctrine/data-fixtures/blob/318df610720a7a7de0add8b27107654877642ac4/lib/Doctrine/Common/DataFixtures/ProxyReferenceRepository.php#L115

There is a reference to LiipFunctionalTestBundle in the documentation of this class: https://github.com/doctrine/data-fixtures/blob/318df610720a7a7de0add8b27107654877642ac4/lib/Doctrine/Common/DataFixtures/ProxyReferenceRepository.php#L115

README.md
```diff
@@ -316,7 +316,18 @@ Tips for Fixture Loading Tests
...
 
- 3. Load your Doctrine fixtures in your tests:
+ 3. For create custom database cache service
+ 
+    **Attention: Don't work with `liip_functional_test.cache_sqlite_db` use only `liip_functional_test.cache_db.*`.**
+    ```yaml
+    # app/config/config_test.yml
+    liip_functional_test:
+        cache_db:
+            sqlite: liip_functional_test.database_backup.sqllite
```
lsmith77: `sqlite` ?

src/DependencyInjection/Configuration.php
```diff
@@ -31,6 +31,15 @@ public function getConfigTreeBuilder(): TreeBuilder
         $rootNode
             ->children()
                 ->booleanNode('cache_sqlite_db')->defaultFalse()->end()
+                ->arrayNode('cache_db')
```
lsmith77: should we then remove `cache_sqlite_db` ?

src/Services/DatabaseBackup/MysqlCustomDatabaseBackup.php
```diff
@@ -0,0 +1,111 @@
...
+        $dbPort = isset($params['port']) ? $params['port'] : '';
+        $dbUser = isset($params['user']) ? $params['user'] : '';
+        $dbPass = isset($params['password']) ? $params['password'] : '';
+
+        $executor->getReferenceRepository()->save($this->getBackupName());
+        /** @var EntityManager $em */
+        $em = $executor->getReferenceRepository()->getManager();
+        self::$metadata = $em->getMetadataFactory()->getLoadedMetadata();
+
+        exec("mysqldump -h $dbHost -u $dbUser -p$dbPass --no-create-db $dbName > {$this->getBackupName()}");
```
lsmith77: https://github.com/liip/LiipFunctionalTestBundle/pull/398#issuecomment-366554219

lsmith77:
didn't do a detailed code review .. but yes splitting this out is very good and fits our goals from https://github.com/liip/LiipFunctionalTestBundle/issues/392

README.md
alekseytupichenkov: Yeah, it's typo

src/DependencyInjection/Configuration.php
alekseytupichenkov: Yes, I'll remove it.
https://github.com/liip/LiipFunctionalTestBundle/pull/398#issuecomment-365875327

src/Services/DatabaseBackup/MysqlCustomDatabaseBackup.php
alekseytupichenkov: @alexislefebvre @lsmith77 
About https://github.com/liip/LiipFunctionalTestBundle/pull/398#issuecomment-366554219, it's not suit, this method save only reference data, but I need save database data (sql with inserts data per each table)

lsmith77: Indeed ..

Generally I think if we have to use a CLI command, then its not really a good fit for this Bundle, since it will not be applicable in too many cases.

This library here might help however https://github.com/portphp/doctrine

alekseytupichenkov: Hmm... I added mysql class just for example
I can use `portphp/doctrine` or something else, it will be interesting for me, but I have fears about performance and additional vendor...

Currently I rename service `mysql_custom` -> `mysql_example` and add attention block from readme

As a proposal, mysql (and other) can be implemented by a separate repo in github, what do you think about it? @alexislefebvre @lsmith77

alexislefebvre: @alekseytupichenkov 

> About #398 (comment), it's not suit, this method save only reference data, but I need save database data (sql with inserts data per each table)

The method works on the current version of this bundle. I'm sorry but I don't understand why you don't use it in this PR. Could you please explain why you chose to perform a full export instead of using only the reference data?

alekseytupichenkov: @alexislefebvre I save reference too
https://github.com/liip/LiipFunctionalTestBundle/pull/398/files#diff-f2e540017ed6e7479f8d4a4b1e17a480R82

But it's not enough, I need backup both, DB and reference.
In case of SQLite, to copy DB enough copy just file, in case of MySql we can't just copy files, need make backup, for make backup I use `mysqldump` because Doctrine can't do it

alekseytupichenkov: @lsmith77 about https://github.com/portphp/doctrine

This lib really cool, but still not suitable, this lib saves only entities, but I need dump full db.
For example, after loading fixtures I can have some additional data, by reason to Doctrine lifecycle methods or DB triggers

I found some PHP libs to make backup like https://github.com/backup-manager/backup-manager or https://github.com/spatie/db-dumper
But they all use `mysqldump` too

alexislefebvre: Thanks for your explanation, I was confused by the fact that SQLite and MySQL need different ways to perform backups. I understand now. Thanks.

alexislefebvre:
It looks like mysql dumps are causing headaches. We may merge this PR without the MySQL backup and add it later in another PR?

Do you agree @alekseytupichenkov? If it's OK for you, you can finalize this PR, move the changes to another branch and open another PR.

powerpd:
@alexislefebvre Yeah, it's ok, I'll make new PR soon

alekseytupichenkov:
Added separate PRs with database cache services (mysql and mongodb)

lsmith77:
only did a very quick review but looks great!

one thing to note in general, we probably should create an UPGRADING.md file to document all the BC breaks and how to migrate existing code to the new version.

lsmith77:
ah .. totally overlooked we already have such a file. thank you for adding .. all good from my side then.

lsmith77:
really awesome work .. and thank you very much for being so receptive on the feedback you got.

alexislefebvre:
Thanks @alekseytupichenkov for the amazing work and @lsmith77 for the review and merge. :tada: 

I published a new release with these changes: https://github.com/liip/LiipFunctionalTestBundle/releases/tag/2.0.0-alpha4

Soullivaneuh:
This PR cause troubles on some configurations. Please see https://github.com/liip/LiipFunctionalTestBundle/issues/423.

PR #402, opened 2018-03-04
database cache from 2.x

alekseytupichenkov:
required https://github.com/liip/LiipFunctionalTestBundle/pull/398

### DIFF
--- a/src/Services/DatabaseBackup/MongodbDatabaseBackup.php
+++ b/src/Services/DatabaseBackup/MongodbDatabaseBackup.php
@@ -1,0 +1,118 @@
+<?php
+
+/*
+ * This file is part of the Liip/FunctionalTestBundle
+ *
+ * (c) Lukas Kahwe Smith <smith@pooteeweet.org>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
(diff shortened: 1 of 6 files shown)


src/Test/WebTestCase.php
```diff
@@ -72,11 +72,6 @@
      */
     private $excludedDoctrineTables = [];
 
-    /**
-     * @var array
-     */
-    private static $cachedMetadatas = [];
```
alexislefebvre: I don't understand why this change is here, since it was removed in your previous PR: https://github.com/liip/LiipFunctionalTestBundle/pull/398/files#diff-3216cb8554ab63f3b7b8531ebcee5156

Maybe rebasing your PR on the `2.x` could fix this issue?

alekseytupichenkov: Hm.. I'll check it today

alekseytupichenkov: Done

src/Services/DatabaseBackup/MongodbDatabaseBackup.php
```diff
@@ -0,0 +1,118 @@
+<?php
+
+/*
+ * This file is part of the Liip/FunctionalTestBundle
+ *
+ * (c) Lukas Kahwe Smith <smith@pooteeweet.org>
```
maximgubar: does the code actually belongs to Lukas ?

PR #408, opened 2018-03-25
[master] Composer: Disallow Symfony 4

alexislefebvre:
README: Suggest to use the last version

In order to fix an issue I introduced: https://github.com/liip/LiipFunctionalTestBundle/pull/351#issuecomment-350216818

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,12 +21,12 @@
         "doctrine/common": "~2.0"
     },
     "require-dev": {
-        "symfony/symfony": "^2.7.1|^3.3|^4.0",
-        "symfony/phpunit-bridge": "^2.7|^3.0|^4.0",
+        "symfony/symfony": "^2.7.1|^3.3",
+        "symfony/phpunit-bridge": "^2.7|^3.0",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -21,12 +21,12 @@
         "doctrine/common": "~2.0"
     },
     "require-dev": {
-        "symfony/symfony": "^2.7.1|^3.3|^4.0",
-        "symfony/phpunit-bridge": "^2.7|^3.0|^4.0",
+        "symfony/symfony": "^2.7.1|^3.3",
```
Jean85: This change is good, but unfortunately we have previous ones that will have those wrong constraints... Can we do nothing about it?

alexislefebvre: You're right. Unfortunately, I don't have any solution for previous versions. We can't delete them since it may break existing tests, CI, etc. It would be better to mark previous releases as deprecated but I don't think that we can do that with Composer.

Jean85: It seems not: https://github.com/composer/composer/issues/7194

PR #410, opened 2018-03-25
[master] Composer: Disallow Symfony 4

alexislefebvre:
Remove lines that should have been removed
in 95b44e969ac3571283d748a5bb217ddd448cb7c2

I forgot some lines in #408…

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ Installation
     following command to download the latest stable version of this bundle:
 
     ```bash
-    $ composer require --dev liip/functional-test-bundle:~1.9.4
+    $ composer require --dev liip/functional-test-bundle:~1.9.5
     ```
 
     This command requires you to have Composer installed globally, as explained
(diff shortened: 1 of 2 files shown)


Soullivaneuh:
Why this remove? Not compatible with Symfony 4 here?

The issue with, people will be allowed to install 1.9.4 with Symfony v4 I'm afraid. :-/

Jean85:
Yes, 1.9.4 was a mistake. This was discussed in composer/composer#7194 too

alexislefebvre:
Yes, the `master` branch is not compatible with Symfony 4.

> The issue with, people will be allowed to install 1.9.4 with Symfony v4 I'm afraid. :-/

That's true but there's nothing we can do, see https://github.com/liip/LiipFunctionalTestBundle/pull/408#discussion_r177685068

That's why the README suggest to require the 1.9.5 version of the bundle that has the right requirements.

Issue #411, opened 2018-04-17
[2.0] Cannot load dependent fixtures that needs dependency injection


magnetik:
Using DoctrineFixtures and LiipFunctionalTestBundle.
 Let's say I've `FooFixtures` that requires `BarFixtures` and `BarFixtures` requires a service injected in it's constructor.

In my test, I use `$this->loadfixtures([FooFictures::class]);`.
This ends up in the `FixturesLoaderFactory` line 65: `$loader->addFixture($fixture);`
This calls the `\Doctrine\Common\DataFixtures\Loader` line 136 that tries to manage dependencies: `elseif ($fixture instanceof DependentFixtureInterface) {  $this->addFixture($this->createFixture($class));`

So if my `BarFixtures` needs a service in the construct, it cannot be instantiated. 

Should the `SymfonyFixturesLoader` from the `DoctrineFixtureBundle` be used?

Thanks

magnetik:
Related: https://github.com/doctrine/DoctrineFixturesBundle/commit/ee5eb61c41572963299c41489b9acc159cbc246d#diff-054595578db32911a949ec0d1754d3c6

Jean85:
As pointed out in https://github.com/liip/LiipFunctionalTestBundle/pull/413#discussion_r182337857, there's a decision to be made in regards of the DoctrineFixtureBundle:

> [...] have to either require it in the composer.json, or make this optional (and add approriate check later in the code).
> 
> I heavily lean on the latter, since requiring an entire Doctrine bundle where it could not be wanted is too hard of a requirement.

alexislefebvre:
There is a conversation about dependencies in a Doctrine Data Fixtures issue: https://github.com/doctrine/data-fixtures/pull/291

magnetik:
To have it working as soon as possible, I'll update my pull request to load the service `SymfonyFixturesLoader` only if it's present, and keep the old behavior otherwise.

Jean85:
Looks like a good solution for me! :+1:

magnetik:
I tried to implement it, but it was.. messy to come up with something working in both old and new case.

If the doctrine fixtures bundle is loaded, then ALL your fixtures have to be services, and be tagged with `doctrine.orm.fixtures` (it's the way the `SymfonyFixturesLoader` works).

The readme states:
> If you plan on loading fixtures with your tests, make sure you have the DoctrineFixturesBundle installed and configured first:

[...]

PR #413, opened 2018-04-18
Quick & ugly fix for being able to load dependent fixtures as service

magnetik:
Contains the tests of #412 and a fix for it. Related to bug #411

It's super ugly and might be broken in every way but makes the test pass  ¯\_(ツ)_/¯.

I feel like relying on `\Doctrine\Bundle\FixturesBundle\Loader\SymfonyFixturesLoader` is a better idea than redoing it in this bundle.

Is the `\Liip\FunctionalTestBundle\Services\FixturesLoaderFactory` required at all? not sure.

Open for discussion :)

### DIFF
--- a/src/Resources/config/database_tools.xml
+++ b/src/Resources/config/database_tools.xml
@@ -7,6 +7,7 @@
     <services>
         <service id="liip_functional_test.services.fixtures_loader_factory" class="Liip\FunctionalTestBundle\Services\FixturesLoaderFactory" public="true">
             <argument type="service" id="service_container" />
+            <argument type="service" id="doctrine.fixtures.loader" />
         </service>
 
         <service id="liip_functional_test.services_database_backup.sqlite" class="Liip\FunctionalTestBundle\Services\DatabaseBackup\SqliteDatabaseBackup" public="true">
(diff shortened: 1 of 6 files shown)


src/Resources/config/database_tools.xml
```diff
@@ -7,6 +7,7 @@
     <services>
         <service id="liip_functional_test.services.fixtures_loader_factory" class="Liip\FunctionalTestBundle\Services\FixturesLoaderFactory" public="true">
             <argument type="service" id="service_container" />
+            <argument type="service" id="doctrine.fixtures.loader" />
```
Jean85: Before adding this hard dependency, you have to either require it in the `composer.json`, or make this optional (and add approriate check later in the code).

I heavily lean on the latter, since requiring an entire Doctrine bundle where it could not be wanted is too hard of a requirement.

magnetik: Thanks for you input. This indeed should be handled differently. 

Should I mimic the behavior of the `\Doctrine\Bundle\FixturesBundle\Loader\SymfonyFixturesLoader` in this bundle directly? 
eg. I get all service tagged with "doctrine.orm.fixtures" and return their instance instead of waiting to create them?

PR #418, opened 2018-05-09
update readme

Issam7:
Hello,
I made this PR to update documentation for supporting sf4, because the old configuration doesn't work for sf4.


README.md
```diff
@@ -329,7 +347,8 @@ Tips for Fixture Loading Tests
     **Attention: you need Doctrine >= 2.2 to use this feature.**
 
     ```yaml
-    # app/config/config_test.yml
+    # sf3: app/config/config_test.yml
+    # sf4: config/packages/test/framework.yaml
     liip_functional_test:
         cache_sqlite_db: true
```
alexislefebvre: This has changed with the 2.x version of this bundle: https://github.com/liip/LiipFunctionalTestBundle/blob/d3f3452aa7c2a82d2316dc0042602a8ba08955ed/UPGRADE-2.0.md

alexislefebvre:
The master branch corresponds to the `1.x` branch, this version will never be compatible with Symfony. So I think that we should document Symfony 4 only on the `2.x` branch.

Could you update your PR so that it targets the `2.x` branch?

Then we'll be able to add a link to the new README in the note about versions added in https://github.com/liip/LiipFunctionalTestBundle/pull/409/files

Issue #394, opened 2018-02-10
Include database creation in a method


alexislefebvre:
A part of the documentation explains how to create a database in `setUp()`: https://github.com/liip/LiipFunctionalTestBundle/blob/25aec2050461bc2ab01dcb1e2619d4259ca0eeda/README.md#non-sqlite

This code should be added in a method in order to remove the need to copy and paste it in every project that uses MySQL (and other databases?).

Jean85:
That would make sense, and obviously we would require the fixtures as an array argument...
It could be helpful with #391, because this method could be the one dedicated to the non-append situation, and instead an other one could be dedicated to the append situation.

alexislefebvre:
It looks like database is already created succesfully: https://github.com/liip/LiipFunctionalTestBundle/blob/b919a90c7c933be7085705e1dd0edffecd25a70f/src/Services/DatabaseTools/ORMDatabaseTool.php#L54

alexislefebvre:
I'm closing this since it's done automatically for SQLite and MySQL, see #426 for reference.

PR #426, opened 2018-05-26
Travis CI: Remove MySQL database creation

alexislefebvre:
This is a test for #394.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,10 +39,6 @@ install:
   # use "update" instead of "install" since it allows using the "--prefer-lowest" option
   - composer update $COMPOSER_FLAGS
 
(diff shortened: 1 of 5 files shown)


alexislefebvre:
This fails in job with `--prefer-lowest` because the kernel is booted before the database is created.

alexislefebvre:
With lowest dependencies, tests are broken. But calling `composer update -vv --profile --no-progress "doctrine/common" "doctrine/dbal" "doctrine/orm"` fixes the issue. I have to find which package require an upgrade, then I'll be able to update `composer.json` in order to require a compatible version.

alexislefebvre:
I upgraded `doctrine-common` in https://github.com/liip/LiipFunctionalTestBundle/pull/426/commits/86fcf6169492cddf92090a15ee50afa779d77d7b but I think that this change is too risky. I'll try to find another way.

PR #429, opened 2018-06-04
Access special test container in getContainer(). Closes #427

### LINKED ISSUE #427: Provide access to special service container in getContainer()
In Symfony 4.1, WebTestCase/KernelTestCase can now access a special container.

You can see in [KernelTestCase:70](https://github.com/symfony/symfony/blob/b51b17d7e8cda2f1b88a4448c5d2e403fd61ad49/src/Symfony/Bundle/FrameworkBundle/Test/KernelTestCase.php#L70), this occurs by checking for the presence of the special container and attempting to use that to set the `static::$container`.

However, calls to `$this->getContainer()` before the Kernel may have been booted results in us creating a kernel and using the standard container [here](https://github.com/liip/LiipFunctionalTestBundle/blob/ad749cd3470e4b2d640d16b3b664ff2f0bb153f5/src/Test/WebTestCase.php#L221)

I think we should do something to achieve the same result so that users of the package can use getContainer() to access the special container to ease in testing.

tomtomau:
See #427 for related discussion.

### DIFF
--- a/src/Test/WebTestCase.php
+++ b/src/Test/WebTestCase.php
@@ -218,7 +218,9 @@ abstract class WebTestCase extends BaseWebTestCase
             $kernel = $this->createKernel($options);
             $kernel->boot();
 
-            $this->containers[$cacheKey] = $kernel->getContainer();
+            $baseContainer = $kernel->getContainer();
+            // Try to check for the special test container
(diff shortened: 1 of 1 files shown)


maximgubar:
merged in #428

PR #432, opened 2018-06-14
Use SymfonyFixturesLoader to load fixtures

### LINKED ISSUE #411: [2.0] Cannot load dependent fixtures that needs dependency injection

magnetik:
It's another try to fix #411.

The downside of the `SymfonyFixturesLoader` is that it always load all fixtures and cannot load a subset of it.

So I wrapped around some code to load only needed dependencies.

The test are not passing, but I'm not sure why as I can use it in my project.


alexislefebvre:
@magnetik I fixed tests in #433, you can rebase your PR on the `2.x` branch and tests should work again.

rvanlaak:
As `"symfony/framework-bundle": "^3.4 || ^4.0"` is the requirement on `2.x`, having auto-wiring on fixtures as the new default would be great for the new release 👍

magnetik:
I would love to see it merged too, but it seems that the bug are failing because of me :)

I'm quite confident that my code is working as I just my fork with a project using symfony 4.1, but the test are failing because I don't understand

magnetik:
I think some test are failing because of my limited knowledge of functional tests. I'll take a look tomorrow but if you feel like taking a look I can use some help ! 

I'm quite confident that the feature is working as I use it in a dev project I have.

tests/App/config.yml
```diff
@@ -77,3 +77,10 @@ security:
             provider: chain_provider
     access_control:
         - { path: ^/, roles: IS_AUTHENTICATED_ANONYMOUSLY }
+
+services:
+    _defaults:
+        autowire: true
+        autoconfigure: true
+    Liip\FunctionalTestBundle\Tests\App\DataFixtures\ORM\:
```
rvanlaak: I think tests are failing because the service on this resource import should be tagged as fixture. 

```
Liip\FunctionalTestBundle\Tests\App\DataFixtures\ORM\:
    resource: 'DataFixtures/ORM/*'
    tags: ['doctrine.fixture.orm']
```

see http://symfony.com/doc/current/bundles/DoctrineFixturesBundle/index.html#loading-fixtures

rvanlaak: created https://github.com/magnetik/LiipFunctionalTestBundle/pull/1 for you

magnetik: humm, I was thinking that the "autoconfigure" should have taken care of this automagically (as test are implementing the right interface)

rvanlaak: I am not sure if DoctrineFixturesBundle did add `__instanceof`, but you might be right. Checking now if they did..

https://symfony.com/blog/new-in-symfony-3-3-service-autoconfiguration

rvanlaak: You are right, implementing `ORMFixtureInterface` should be enough to autoconfigure the `doctrine.fixture.orm` tag, whereafter `FixturesCompilerPass` will add the fixture to the definition of `doctrine.fixtures.loader`.

* https://github.com/doctrine/DoctrineFixturesBundle/blob/master/DependencyInjection/DoctrineFixturesExtension.php
* https://github.com/doctrine/DoctrineFixturesBundle/blob/master/DependencyInjection/CompilerPass/FixturesCompilerPass.php

magnetik: I've added the tag definition in the `config.yml` and the test are passing. Is that something that can be related to order in which the services are loaded ?

rvanlaak: Then I guess the `DoctrineFixturesBundle` is not loaded / included with the `test` environment's `AppKernel`? 

That would also explain why it's working in your dev environment, as you probably did include the bundle in your kernel over there.

magnetik:
Okay so it requires `doctrine-fixtures-bundle` 3.0.2 (3.0.0 is bugged). 

I've adapted the composer.json :)

rvanlaak:
🎉 

@alexislefebvre would be great if `2.0.0-alpha5` could get released after this PR is merged

rvanlaak:
Or maybe @maximgubar could help us in getting this merged and releasing `2.0.0-alpha5`?

tests/App/DataFixtures/ORM/LoadUserWithServiceData.php
```diff
@@ -0,0 +1,58 @@
...
+use Liip\FunctionalTestBundle\Tests\App\Entity\User;
+use Symfony\Component\Security\Csrf\TokenStorage\TokenStorageInterface;
+
+class LoadUserWithServiceData extends AbstractFixture implements FixtureInterface
+{
+    private $tokenStorage;
+
+    public function __construct(TokenStorageInterface $tokenStorage)
+    {
+        $this->tokenStorage = $tokenStorage;
```
alexislefebvre: Is there a way to use `$this->tokenStorage` in this test? I would like to ensure that this service is callable.

magnetik: I've added a call in the `load` method.

alexislefebvre:
LGTM, I'm surprised that it doesn't break anything even though the fixture loader has been changed.

Thanks!

magnetik:
I'm expecting quite a few bugs to pop in, regarding DX with the declaration of the service and so one. 

Let's see how it goes when an alpha is released :)

rvanlaak:
Where the previous fixture loader did simply rely on plain old class instantiation or a service definition, the loader now basically is enhanced by one that supports autowiring and autoconfiguration. In other words, as long as userland's service definitions are valid everything should be fine.  [Before](https://github.com/liip/LiipFunctionalTestBundle/blob/2.x/src/Services/FixturesLoaderFactory.php#L55) vs [After](https://github.com/liip/LiipFunctionalTestBundle/pull/432/files#diff-cab69af81212168b9512d403b080e023R19) 🎉 Really looking forward to this!

alexislefebvre:
Could you please add a small documentation about these changes in https://github.com/liip/LiipFunctionalTestBundle/blob/2.x/CHANGELOG.md, and https://github.com/liip/LiipFunctionalTestBundle/blob/2.x/UPGRADE-2.0.md if this is a breaking change?

magnetik:
Done

alexislefebvre:
Thanks @magnetik and @rvanlaak for your work on this PR, I released the 2.0.0-alpha5 version: https://github.com/liip/LiipFunctionalTestBundle/releases/tag/2.0.0-alpha5

composer.json
```diff
@@ -25,9 +25,9 @@
         "symfony/phpunit-bridge": "^3.4 || ^4.0",
         "doctrine/orm": "^2.6",
         "symfony/monolog-bundle": "^3.1",
-        "doctrine/doctrine-fixtures-bundle": "^3.0",
+        "doctrine/doctrine-fixtures-bundle": "^3.0 >=3.0.2",
```
Jibbarth: This line shouldn't be in `require` section instead of `require-dev` ? 
Since we use it in `Services\FixturesLoaderFactory`, it's became a dependency, WDYT ?

Got this error when updating to 2.0.0-alpha5

```
The service "liip_functional_test.services.fixtures_loader_factory" has a dependency on a non-existent service "doctrine.fixtures.loader". 
```

magnetik: see https://github.com/liip/LiipFunctionalTestBundle/pull/437

PR #435, opened 2018-07-17
Remove double copyright header

magnetik:


Jean85:
Shouldn't we fix this stuff with a CS linter?

PR #436, opened 2018-07-17
Make some class final

### LINKED ISSUE #380: Make classes final?
We should take benefit of the next major release to make most of the classes final.

magnetik:
Hello
I did not set final on the `Liip\FunctionalTestBundle\Services\DatabaseTools` because we already extends ORM for Sqlite so it feels like user might need it too.

Maybe we should set more class final?

Fixes #380

### DIFF
--- a/src/Annotations/QueryCount.php
+++ b/src/Annotations/QueryCount.php
@@ -17,7 +17,7 @@ namespace Liip\FunctionalTestBundle\Annotations;
  * @Annotation
  * @Target({"METHOD"})
  */
-class QueryCount
+final class QueryCount
 {
     /** @var int */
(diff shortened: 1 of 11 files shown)


alexislefebvre:
Marking `*DatabaseBackup` classes as `final` may prevent to creating classes that extend them and override default services: https://github.com/liip/LiipFunctionalTestBundle/blob/a0307d1656343d4ab3e11712408d07e94b731238/src/Resources/config/database_tools.xml#L18

But we'll be able to remove these `keywords` at any time if needed: https://github.com/liip/LiipFunctionalTestBundle/issues/380#issuecomment-358304972

So we can merge this, thanks.

PR #437, opened 2018-07-17
Make doctrine-fixtures-bundle non mandatory

magnetik:
After this comment: https://github.com/liip/LiipFunctionalTestBundle/pull/432#discussion_r202950021

It does not looks like the best way to do it, but all database tools currently requires the `fixtures_loader_factory` service.


alexislefebvre:
Thanks all for your multiple PRs, reviews and more. Here is the last release: https://github.com/liip/LiipFunctionalTestBundle/releases/tag/2.0.0-alpha6

Issue #434, opened 2018-07-13
Doctrine\Common\ClassLoader is deprecated.


freedelta:
### Preconditions

1. This error is fired when launching tests in symfony: **Doctrine\Common\ClassLoader** is deprecated
2. The error comes with Doctrine Common v2.9.0
3. It seems that the ClassLoader in doctrine common is deprecated
https://github.com/doctrine/common/issues/826

### Steps to reproduce

 I think the error comes from the fact that you require doctrine/common in your composer.json

### Expected result

Should not use doctrine/common

### Actual result

In our logs:  **Doctrine\Common\ClassLoader** is deprecated

alexislefebvre:
Thanks for your report. Did you see this deprecation with the `1.x` or `2.x` version of this bundle?

nataliastanko:
> Thanks for your report. Did you see this deprecation with the 1.x or 2.x version of this bundle?

I can see it using doctrine/common 2.9.x-dev

alexislefebvre:
I removed the direct dependency in #440 and 3 of the dependencies of this bundle still require `doctrine/common`:

```
$ composer why doctrine/common
doctrine/data-fixtures  v1.3.1  requires  doctrine/common (~2.2)         
doctrine/orm            v2.6.2  requires  doctrine/common (^2.7.1)       
symfony/symfony         v4.1.1  requires  doctrine/common (~2.4@stable)  
```

So, we'll see the deprecation message until these 3 packages remove `doctrine/common` from their dependencies.

---

We can do the same thing as in https://github.com/symfony/symfony/pull/27609 and require the necessary doctrine components and drop `doctrine/common`.

alexislefebvre:
As noted here https://github.com/symfony/symfony/pull/27609#issuecomment-405472257, the parameter [`weak_vendors`](https://symfony.com/doc/3.4/components/phpunit_bridge.html#making-tests-fail) can be used to ignore warnings from `vendor/`. I didn't tested it but it should hide Doctrine deprecations until all the packages are updated to remove this warning.

Tobion:
This is not caused by LiipFunctionalTestBundle and should get fixed automatically when doctrine/orm 2.6.3 is released, see https://github.com/doctrine/doctrine2/pull/7307

Tobion:
Reopening as the dependency on doctrine/common should be split like in https://github.com/symfony/symfony/pull/27609

Jean85:
[...]

PR #440, opened 2018-07-17
[2.x] Remove hard dependency on doctrine/common

alexislefebvre:
Related to #434, I think that we don't need this hard dependency.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
     ],
     "require": {
         "php": "^7.1",
-        "doctrine/common": "^2.0",
+        "doctrine/annotations": "^1.3",
         "symfony/browser-kit": "^3.4 || ^4.1",
         "symfony/framework-bundle": "^3.4 || ^4.1"
     },


Jean85:
Isn't it used anywhere in the code?

alexislefebvre:
I remember it only now, I think that it's used to read annotations in https://github.com/liip/LiipFunctionalTestBundle/blob/294f8127df03923a16354e6f1062f5172be06994/src/Annotations/QueryCount.php#L17-L18

So removing `doctrine/common` seems wrong…

Jean85:
Well, that highlights that we do not have a test in place, probably...

alexislefebvre:
`doctrine/common` was installed on Travis CI after I removed it: https://travis-ci.org/liip/LiipFunctionalTestBundle/jobs/405112576

So it must be a dependency of one of these dependencies: https://github.com/alexislefebvre/LiipFunctionalTestBundle/blob/d1335ad3a37c970a152b533e78ba091fa6370c81/composer.json#L18-L33

`"doctrine/orm": "^2.6",` requires `doctrine/common` so it will always be installed with current versions of Doctrine ORM.

If we don't install `doctrine/orm`, on the CI, we won't be able to count number of queries, I don't know how we can test the state where `doctrine/common` is not installed but we can still run internal tests.

Jean85:
Common is a dependency of Doctrine 2.6, but I think it will be dropped in 2.7, so we cannot rely on that...

alexislefebvre:
We may rely on https://github.com/doctrine/annotations if we only used `doctrine/common` to read annotations.

alexislefebvre:
See https://github.com/liip/LiipFunctionalTestBundle/issues/434#issuecomment-406899337: we have to require the new doctrine packages that provide features from `doctrine/common` that this bundle uses.

Majkl578:
> Common is a dependency of Doctrine 2.6, but I think it will be dropped in 2.7, so we cannot rely on that...

It won't, only in ORM 3.0 (due to dependency on doctrine/common's proxy generator). But it'd be a good move to depend on specific doctrine/* packages (doctrine/annotations in your case) instead of doctrine/common.

General rule of thumb: never rely on indirect dependencies - they may be dropped or bumped to a new major version without you noticing (since you have no constraint against the indirect dependency). 👍 

Also I recommend https://github.com/maglnet/ComposerRequireChecker, it'll spot dependencies that are not directly depended upon.

PR #445, opened 2018-08-07
Allow static assertion calls

scaytrase:
Currently all phpunit assertions are static, which allow call them like `self::assertNull`. Given asserts are not static (but nothing prevents them to be static), so currently we have to write it like

```php
self::assertNotNull($someting);
$this->assertStatusCode(200, $client);
```

Which makes confusion of code style. The given patch would allow every one to choose assertion invokation style

see http://phpunit.readthedocs.io/en/7.1/assertions.html

> A common question, especially from developers new to PHPUnit, is whether using $this->assertTrue() or self::assertTrue(), for instance, is “the right way” to invoke an assertion. The short answer is: there is no right way. And there is no wrong way, either. It is a matter of personal preference.

### DIFF
--- a/Test/WebTestCase.php
+++ b/Test/WebTestCase.php
@@ -76,6 +76,31 @@ abstract class WebTestCase extends BaseWebTestCase
      */
     private static $cachedMetadatas = [];
 
+    /**
+     * Asserts that the HTTP response code of the last request performed by
+     * $client matches the expected code. If not, raises an error with more
+     * information.
+     *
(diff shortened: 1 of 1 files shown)


alexislefebvre:
We will be able to add this to the `2.x` branch.

scaytrase:
All phpunit assertions are static, so this worked for years (as it has been documented already) . 2.x looks good, hope it will be released soon

PR #451, opened 2018-09-25
[2.x] Remove useless check in `WebTestCase::fetchContent`

magnetik:
Given that `$success` is typed as bool, I don't see any usefulness in checking if it's a boolean.


alexislefebvre:
This is probably a check that was added before functions arguments were type-hinted. Thanks!

PR #454, opened 2018-10-02
Fix typo in HttpAssertions::assertValidationErrors

magnetik:
There was only a placeholder for the method and not the class plus they were not in a logical order.


alexislefebvre:
Good catch, thanks! Could you please fix the tests?

PR #459, opened 2018-10-18
Order composer dependencies by name

magnetik:
It's a detail but it helps seeing if a package is there or not.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,30 +16,30 @@
     ],
     "require": {
         "php": "^7.1",
-        "symfony/framework-bundle": "^3.4 || ^4.1",
+        "doctrine/common": "^2.0",
         "symfony/browser-kit": "^3.4 || ^4.1",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -16,30 +16,30 @@
...
-        "twig/twig": "^2.0",
+        "symfony/monolog-bundle": "^3.1",
+        "symfony/phpunit-bridge": "^3.4 || ^4.1",
+        "symfony/symfony": "^3.4 || ^4.1",
         "theofidry/alice-data-fixtures": "^1.0.1",
-        "jackalope/jackalope-doctrine-dbal": "^1.3",
-        "doctrine/data-fixtures": "^1.3"
+        "twig/twig": "^2.0"
     },
     "suggest": {
```
Jean85: We need to commit the actual option:
```suggestion
    "config": {
        "sort-packages": true
    },
    "suggest": {
```

Issue #329, opened 2017-07-23
Getting exit code from runCommand()


IwfY:
For some command testing I want to evaluate the exit code returned by it. With the current implementation of runCommand() in WebTestCase this is not possible.

So I basically copied the function and used a reference parameter to pass the command's exit code.

    /**
     * Builds up the environment to run the given command.
     *
     * alternative to self::runCommand
     *
     * $reuseKernel defaults to true compared to Liip/WebTestCase
     *
     * @param string $name
     * @param array  $params
     * @param bool   $reuseKernel
     * @param int    $runExitCode pointer to be filled with exit code of run command
     *
     * @return string
     */
    protected function runCommandWithExitCode($name, array $params = array(), $reuseKernel = true, &$runExitCode)
    {
        array_unshift($params, $name);

        if (!$reuseKernel) {
            if (null !== static::$kernel) {
                static::$kernel->shutdown();
            }

            $kernel = static::$kernel = $this->createKernel(array('environment' => $this->environment));
            $kernel->boot();
        } else {
            $kernel = $this->getContainer()->get('kernel');
        }

        $application = new Application($kernel);
        $application->setAutoExit(false);

        // @codeCoverageIgnoreStart
        if ('203' === substr(Kernel::VERSION_ID, 0, 3)) {
            $params = $this->configureVerbosityForSymfony203($params);
        }
        // @codeCoverageIgnoreEnd

        $input = new ArrayInput($params);
        $input->setInteractive(false);

        $fp = fopen('php://temp/maxmemory:' . $this->maxMemory, 'r+');
        $output = new StreamOutput($fp, $this->getVerbosityLevel(), $this->getDecorated());

        $runExitCode = $application->run($input, $output);

        rewind($fp);

        return stream_get_contents($fp);
    }

I'd be willing to create a pull request for such functionality if it is wanted in this bundle. I just need some advice how to integrate this feature. Expanding the interface of the current runCommand in the way seen above might not be wanted as it would break backwards compatibility. But having this much duplicate code feels bad too.

Any interest in the feature or preferences how you'd like to have it implemented?

alexislefebvre:
We may add a variable in order to store the value:

```php
$this->lastExitCode = $application->run($input, $output);
```

And a method in order to check the exit code:

```php
public function assertSameLastExitCode($expectedExitCode) {
    // the following code has to be rewritten
    PHPUnit::assertSame($expectedExitCode, $this->lastExitCode);
}
```

[...]

Issue #219, opened 2016-01-14
runCommand(): use CommandTester (Enhancement)


Aerendir:
I'm trying to use the `WebTestCase::runCommand()` method.

I have yet implemented my unit test to test command, but, as I've anyway extend `WebTestBundle` to load fixtures, I like to also use other useful methods that the bundle provides.
So II'd like to rewrite my tests to use those methods.

Now I'm working on a command and seen the code of the method `WebTestCase::runCommand()`.

I noted it doesn't use the helpful `Symfony\Component\Console\Tester\CommandTester()` class: is there a reason for this? Or may be the `CommandTester` class were released after the `WebTestCase::runCommand()`?

Can someone help me understand?

lsmith77:
yes ``WebTestCase::runCommand()`` was created years before ``Symfony\Component\Console\Tester\CommandTester()``

Aerendir:
I'll work on it ASAP...

giosh94mhz:
This would be a cool addition, since with the current `runCommand` there is no status code returned.

Here's what I did:

```php
    $tester = new CommandTester($this->command);
    $this->assertEquals($this->command->getName(), 'my:name');
    $this->assertEquals(0, $tester->execute());
    $output = $tester->getOutput();
```

magnetik:
I've gave it a go in the PR above.

martiis:
I recently updated and got some issues. CommandTester does not dispatch console events and some tests started failing. Is there a solution?

Jean85:
Are you sure that you're not referencing a different kernel?

PR #460, opened 2018-10-18
Use symfony command tester

magnetik:
Relates to #329 and #219

I haven't fixed all tests but just one to verify that it's working.

Should I go forward and fix them all? Is this something wanted?

### DIFF
--- a/src/Test/WebTestCase.php
+++ b/src/Test/WebTestCase.php
@@ -22,8 +22,7 @@ use Symfony\Bundle\FrameworkBundle\Client;
 use Symfony\Bundle\FrameworkBundle\Console\Application;
 use Symfony\Bundle\FrameworkBundle\Test\WebTestCase as BaseWebTestCase;
 use Symfony\Component\BrowserKit\Cookie;
-use Symfony\Component\Console\Input\ArrayInput;
-use Symfony\Component\Console\Output\StreamOutput;
+use Symfony\Component\Console\Tester\CommandTester;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 use Symfony\Component\DependencyInjection\ResettableContainerInterface;
 use Symfony\Component\DomCrawler\Crawler;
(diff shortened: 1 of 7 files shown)


alexislefebvre:
I like this change. Yes please update all the tests.

magnetik:
All 💚.
I've added a little something in the `UPGRADE` and `CHANGELOG`

tests/Command/CommandTest.php
```diff
@@ -27,54 +28,61 @@ class CommandTest extends WebTestCase
...
+        $this->assertContains('Environment: test', $this->commandTester->getDisplay());
+        $this->assertContains('Verbosity level: NORMAL', $this->commandTester->getDisplay());
     }
 
     public function testRunCommandWithoutOptionsAndNotReuseKernel(): void
     {
         // Run command without options
-        $this->display = $this->runCommand('liipfunctionaltestbundle:test');
+        $this->commandTester = $this->runCommand('liipfunctionaltestbundle:test');
+
```
alexislefebvre: We can reduce calls to `getDisplay()` by defining `$this->display` like before:
```suggestion
$this->display =  $this->commandTester->getDisplay();
```

magnetik: @alexislefebvre can we re-have your input on it? 
(About having to use `$this->display` or just a local variable `$display`. Because having only `$display` would change all lines too. )

alexislefebvre: I was thinking about using `$this->display` in the test like before, to reduce the number of changed lines. We can live with this. :+1:

src/Test/WebTestCase.php
```diff
@@ -87,38 +86,33 @@ protected function getServiceMockBuilder(string $id): MockBuilder
...
-
-        $application->run($input, $output);
-
-        rewind($fp);
+        $command = $application->find($name);
+        $commandTester = new CommandTester($command);
+        $commandTester->execute(
+            array_merge(['command' => $command->getName()], $params),
+            [
+                'interative' => false,
```
alexislefebvre: Should it be:
```suggestion
                'interactive' => false,
```
?

magnetik: Indeed, done !

PR #465, opened 2018-10-31
Update docs for Symfony 4 sqlite database setup

slipo:
Here's an update to the docs based on my experience setting this up to use SQLite on Symfony 4 with a default project-wide `doctrine.yaml` configured to use PostgreSQL.

The correct parameter to set the location of the SQLite file is `path`. I'm setting `url` to override any other values for this setting. If you don't set this to `null` and in your main `doctrine.yaml` file you have it set to a database url, Doctrine will give you a cryptic error. Lastly, `driver` is required.

### DIFF
--- a/doc/database.md
+++ b/doc/database.md
@@ -40,7 +40,9 @@ Tips for Fixture Loading Tests
         # config/packages/test/doctrine.yaml
         doctrine:
             dbal:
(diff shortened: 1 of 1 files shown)


alexislefebvre:
Thanks!

PR #379, opened 2018-01-13
Dependencies upgrade

Soullivaneuh:
Closes #268
Closes #297 
Closes #321
Closes #337 (replace)
Closes #366 (replace)
Closes #378 (replace)
Closes #353

Because some dependencies are linked together (e.g. PHPUnit with Symfony, Nelmio alice with data-fixtures...), I decided to make a global PR to update them all.

- [x] Travis update
- [x] Composer dependencies update
- [x] Composer lint (`~` and simple `|` should be used by convention)
- [x] Fix tests
- [x] Rollback `symfony/symfony` dependencies to `^3.4 || ^4.0` when tests will be fixed
- [x] Remove all BC tricks
- [x] PHP migration on php-cs-fixer
- [x] Introduce header comment on php-cs-fixer (to make it working well with strict_type declaration placement)
- [x] Add strict typing on methods

PR #371, opened 2018-01-10
FlintCI integration for 2.x

Soullivaneuh:
See https://github.com/liip/LiipFunctionalTestBundle/pull/370#issuecomment-356691876

PR #467, opened 2018-11-15
[2.0] Fix runCommand PHPDoc

Jean85:
This BC was introduced but hidden to static analyzers due to this PHPDoc error.

Maybe we should introduce static analysis in 2.0 CI?

### DIFF
--- a/src/Test/WebTestCase.php
+++ b/src/Test/WebTestCase.php
@@ -84,7 +84,7 @@ abstract class WebTestCase extends BaseWebTestCase
      * @param array  $params
      * @param bool   $reuseKernel
      *
-     * @return string
+     * @return CommandTester
      */
     protected function runCommand(string $name, array $params = [], bool $reuseKernel = false): CommandTester
     {


alexislefebvre:
Yes, adding static analysis tools is a great idea. FlintCI was added in #371 but I don't know why it doesn't work anymore (this PR has no check from FlintCI).

About this line, the whole PHPDoc block is redundant for this method, maybe we should drop it when it adds no value to the args and return type hints.

Jean85:
* FlintCI is not working and neither Travis, we're out of CI checks! :disappointed: It seems to have stopped working evend mid-PR in #379
 * superfluous PHPDoc could be suppressed using PHP-CS-Fixer, with the `` rule
 * on that regard, I don't see PHP-CS-Fixer required in dev or even run in CI; I see that we have StyleCI, but the config is out of sync with the committed one, that's not good!
 * I can introduce PHPStan in the CI checks, I'll work on a PR ASAP; I think we should move the CS check there too, leveraging Travis' stages

alexislefebvre:
Great news Jean, see #468.

Issue #415, opened 2018-05-03
Backup is wrong


eugenekurasov:
Hi,
when we load fixtures first time - they are saved to backup files with extension .ser
Here is example code where I am expected to have 3 backup files but got only 2.

```
$this->loadFixtures([SomeFixture1::class, SomeFixture2::class], $append = false);
.....
$this->loadFixtures([SomeFixture3::class], $append = true);
....
// This is fixture will be load from backup with data from SomeFixture1, SomeFixture2
$this->loadFixtures([SomeFixture3::class], $append = false);
```

Can I disable loading fixtures from cache?

alexislefebvre:
Hi,
do you use the version 1 or 2 of this bundle? You mention a problem with number of backup files, but are your fixtures loaded correctly?

alexislefebvre:
> Can I disable loading fixtures from cache?

Yes, this is the default behaviour of the bundle.

Did you set `liip_functional_test.cache_sqlite_db: true` in your configuration?

eugenekurasov:
```liip/functional-test-bundle 2.0.0-alpha4``` 

Fixture loaded not corrected. From my example: ```$this->loadFixtures([SomeFixture3::class], $append = false);``` I expect data only from SomeFixture3

Yes I set ```liip_functional_test.cache_sqlite_db: true```.

**Comment by alexislefebvre** [comment]
My mistake, the new parameter for caching database is this one:

```
liip_functional_test:
        cache_db:
            sqlite: liip_functional_test.services_database_backup.sqlite
```

Source: https://github.com/liip/LiipFunctionalTestBundle/blob/d3f3452aa7c2a82d2316dc0042602a8ba08955ed/UPGRADE-2.0.md

eugenekurasov:
@alexislefebvre thanks.

But if cache is on, backup is wrong. You can [see](https://github.com/liip/LiipFunctionalTestBundle/blob/2.0.0-alpha4/src/Services/DatabaseTools/ORMSqliteDatabaseTool.php#L39) that append option does not use in backup service.

eugenekurasov:
@alexislefebvre thank you for work around.

I rewrited test to avoid this situation, so I hope someone going to fix it.

alekseytupichenkov:
I'll fix it soon

alekseytupichenkov:
I'll fix it in PR #462

alexislefebvre:
[...]

PR #476, opened 2018-12-16
2.x: Fix issues with MySQL database backup

alexislefebvre:
See #415

### DIFF
--- a/src/Services/DatabaseBackup/AbstractDatabaseBackup.php
+++ b/src/Services/DatabaseBackup/AbstractDatabaseBackup.php
@@ -33,7 +33,7 @@ abstract class AbstractDatabaseBackup implements DatabaseBackupInterface
      *
      * @var array
      */
-    protected $classNames;
+    protected $classNames = [];
 
     public function __construct(ContainerInterface $container, FixturesLoaderFactory $fixturesLoaderFactory)
     {
@@ -41,10 +41,15 @@ abstract class AbstractDatabaseBackup implements DatabaseBackupInterface
         $this->fixturesLoaderFactory = $fixturesLoaderFactory;
     }
 
-    public function init(array $metadatas, array $classNames): void
+    public function init(array $metadatas, array $classNames, bool $append = false): void
     {
         $this->metadatas = $metadatas;
-        $this->classNames = $classNames;
+
+        if ($append) {
+            $this->classNames = array_merge($this->classNames, $classNames);
+        } else {
+            $this->classNames = $classNames;
+        }
     }
 
     /**
--- a/src/Services/DatabaseBackup/MysqlDatabaseBackup.php
+++ b/src/Services/DatabaseBackup/MysqlDatabaseBackup.php
@@ -94,7 +94,7 @@ final class MysqlDatabaseBackup extends AbstractDatabaseBackup implements Databa
         }
     }
 
-    public function restore(AbstractExecutor $executor): void
+    public function restore(AbstractExecutor $executor, array $excludedTables = []): void
     {
         /** @var EntityManager $em */
         $em = $executor->getReferenceRepository()->getManager();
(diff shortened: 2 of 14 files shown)


alexislefebvre:
Some tests fail, if I run them separately then they pass. I'm wondering how could this happen.

alexislefebvre:
I found a quick and dirty workaround:  Don't use backup service when appending data, see https://github.com/liip/LiipFunctionalTestBundle/pull/476/commits/dec6f27d05798745e03566e3fb0d5be154078ff4

alexislefebvre:
I think that the backup service should keep an history of loaded fixtures so that if we load fixtures `A` then load fixtures `B` with `append = true`, the backup service recognizes that it can loads a backup with `A + B`. But I don't know if it's a good idea and if it's feasible.

src/Services/DatabaseBackup/AbstractDatabaseBackup.php
```diff
@@ -33,18 +33,23 @@
...
     }
 
-    public function init(array $metadatas, array $classNames): void
+    public function init(array $metadatas, array $classNames, bool $append = false): void
     {
         $this->metadatas = $metadatas;
-        $this->classNames = $classNames;
+
+        if ($append) {
+            $this->classNames = array_merge($this->classNames, $classNames);
```
alexislefebvre: When appending fixtures, keep track of `classNames` which had been already loaded.

src/Services/DatabaseBackup/MysqlDatabaseBackup.php
```diff
@@ -84,7 +78,7 @@ public function backup(AbstractExecutor $executor): void
         $executor->getReferenceRepository()->save($this->getBackupFilePath());
         self::$metadata = $em->getMetadataFactory()->getLoadedMetadata();
 
-        exec("mysqldump -h $dbHost -u $dbUser -p$dbPass --no-create-info --skip-triggers --no-create-db --no-tablespaces --compact $dbName > {$this->getBackupFilePath()}");
+        exec("MYSQL_PWD=$dbPass mysqldump --host $dbHost --port=$dbPort --user $dbUser --no-create-info --skip-triggers --no-create-db --no-tablespaces --compact $dbName > {$this->getBackupFilePath()}");
```
alexislefebvre: Using an environment variable remove this error: `Warning: Using a password on the command line interface can be insecure.`.

tests/AppConfigMysqlCacheDb/AppConfigMysqlKernelCacheDb.php
```diff
@@ -0,0 +1,41 @@
...
+ * (c) Lukas Kahwe Smith <smith@pooteeweet.org>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Liip\FunctionalTestBundle\Tests\AppConfigMysqlCacheDb;
+
+/*
+ * This file is part of the Liip/FunctionalTestBundle
```
magnetik: Double header

alexislefebvre:
@magnetik Yes we could add an annotation to disable the cache for a test.

alexislefebvre:
@magnetik I created an issue with your idea: #480

VincentClair:
Cache seems not be compatible with loadFixtureFiles().
The backup file is empty and $classNames argument of loadFixtures() empty too.
I have missconfigure something ?

alexislefebvre:
If you upgraded the bundle, the old cache can cause issues. You can remove it and try again.

VincentClair:
It's not a Symfony cache problem, I already drop completely the folder ;)

After investigation the code of the last alpha release, loadFixtureFiles() not used the cache system:
WebTestCase::loadFixtureFiles() calls AbstractDatabaseTool::loadAliceFixture(). Inside latter is called cleanDatabase() without parameters, that causing a empty cache file for database. Then fixtures are loaded from files thanks to fidry_alice_data_fixtures.loader.doctrine service.

Maybe there could be a cache for this external service from theofidry/alice-data-fixtures library to activate ?

PR #486, opened 2019-01-16
Add php7.3 in travis tests

magnetik:
Not sure about adding the same `EXTRA_PACKAGES` as the one with php 7.2.
Maybe a comment should be added on why this is necessary.


magnetik:
Why isn't travis building this ? 🤔

alexislefebvre:
Travis CI didn't trigger a build because of an error in the YAML file :wink:

PR #491, opened 2019-02-01
Fix #273: form validation works with sub-requests

### LINKED ISSUE #273: assertValidationErrors issue if using twig {{ render(controller(...)) }}
All the `assertValidationErrors` statements expecting validation errors fail with an empty actual list of validation errors when the page renders another controller with twig `{{ render(controller() }}` tag, even with a dummy controller with empty view.

Tested with Symfony 2.8 and LiipFunctionalTestBundle 1.6.3

AkenRoberts:
Since the validation service is reused (as a "singleton" if you will) throughout the testing process, any validation errors are cleared for each request via the `kernel.request` event.

When an embedded controller is used, a new sub-request is created which will fire off another `kernel.request` event. This clears any form errors _during_ the test, and causes `assertValidationErrors` to fail incorrectly.

By checking if the event is the master request, we can allow embedded controllers.

I'm not sure if this will impact the possible use of forms within a sub-request. I don't know if that's a thing. 🤔

Thank you for the bundle. 😄

### DIFF
--- a/src/Validator/DataCollectingValidator.php
+++ b/src/Validator/DataCollectingValidator.php
@@ -14,6 +14,7 @@ declare(strict_types=1);
 namespace Liip\FunctionalTestBundle\Validator;
 
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpKernel\Event\GetResponseEvent;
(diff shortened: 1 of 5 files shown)


alexislefebvre:
See #273

Issue #470, opened 2018-11-29
Fails to connect into mysql database when database not created


webmake:
Hello,

This library fails to connect into mysql database when database not created with error `An exception occurred in driver: SQLSTATE[HY000] [1049] Unknown database 'test'`

Place where error occurs:
https://github.com/liip/LiipFunctionalTestBundle/blob/2.0.0-alpha10/src/Services/DatabaseTools/ORMDatabaseTool.php#L63 or [commit to be exact](https://github.com/liip/LiipFunctionalTestBundle/blob/ee1b91e16ab20286503051d3478b6438287c36e4/src/Services/DatabaseTools/ORMDatabaseTool.php#L63) fails because used dbal library performs connect https://github.com/doctrine/dbal/blob/b45ed5e13e4834f64fbc56bfdc0344430cecae63/lib/Doctrine/DBAL/Connection.php#L358 into database despite `unset($params['dbname']);` in your class [`ORMDatabaseTool`](https://github.com/liip/LiipFunctionalTestBundle/blob/2.0.0-alpha10/src/Services/DatabaseTools/ORMDatabaseTool.php#L61) in 61 line

### Steps to reproduce
```
class MyClassTest extends \Liip\FunctionalTestBundle\Test\WebTestCase
{
    protected function setUp(): void
    {
        self::bootKernel();
        $this->loadFixtures([
            LoadEntities::class,
        ]);
        parent::setUp();
    }
}
```
and 
```
"doctrine/dbal": "^2.8",
"liip/functional-test-bundle": "~2.0@alpha",
```

alexislefebvre:
As a workaround, you can create the database before running tests:

```
(app|bin)/console doctrine:database:create --env=test --if-not-exists

**Comment by alexislefebvre** [comment]
I can't reproduce this bug on my local environment.

Do you use `driver: pdo_mysql` in your configuration?

**Comment by alexislefebvre** [comment]
I changed the database name in the configuration used on Travis CI and the tests passed: #477

It looks like a specific configuration leads to the error you have, but I'm still wondering what are these parameters.

**Comment by webmake** [comment]
```
parameters:
    env(DATABASE_URL): ''
    env(DATABASE_SETTINGS_URL): ''

doctrine:
    dbal:
        default_connection: default
        connections:
            default:
                driver: 'pdo_mysql'
                server_version: '5.7'
                charset: utf8mb4
                default_table_options:
                    charset: utf8mb4
[...]

PR #494, opened 2019-03-02
2.x: Add tests with fixtures loading in setUp

alexislefebvre:


alexislefebvre:
See #470

alexislefebvre:
With database cache, users loaded from backup have ids from 3 to 12 instead of 1 to 10, this breaks tests.

alexislefebvre:
I had to reset `AUTO_INCREMENT` so that the database cache load data with the same ids than before, but I'm not sure if this change is a good idea…

dbu:
i guess the tests relying on a specific autoincrement are a bit of a dangerous idea. but looking at the [documentation](https://oracle-base.com/articles/mysql/mysql-how-truncate-table-affects-auto-increment), truncate did reset the autoincrement, so i think its good to reset it.

PR #500, opened 2019-03-21
WebTestCase makeClient method update

DamienVauchel:
Hi,

I detected an useless if condition on authentication so I removed it.

### DIFF
--- a/src/Test/WebTestCase.php
+++ b/src/Test/WebTestCase.php
@@ -363,15 +363,13 @@ abstract class WebTestCase extends BaseWebTestCase
      */
     protected function makeClient($authentication = false, array $params = []): Client
     {
-        if ($authentication) {
-            if (true === $authentication) {
-                $authentication = [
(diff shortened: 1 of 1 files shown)


alexislefebvre:
Thanks, with this change we can't pass an array as the first parameter. That's why there are 2 conditions. It looks like it's a breaking change and it should break internal tests of the bundle.

DamienVauchel:
My bad! I didn't check it.

Don't you think it is kinda dirty to accept that an array argument could be a boolean (or a boolean argument be an array)? 
(Not a mean/troll question, I don't know if my english is perfect^^", it is a real question)

alexislefebvre:
I agree with you, the fact that the first parameter accept two different types is confusing.

We should refactor this for the 3.x release, see #496 

We may add a `makeAuthentifiedClient()` method in order to reduce the size and complexity of `makeClient()`.

alexislefebvre:
See #506.

PR #506, opened 2019-04-24
3.x: extract code from makeClient to 2 new methods

alexislefebvre:
Replaces #500

### DIFF
--- a/src/Test/WebTestCase.php
+++ b/src/Test/WebTestCase.php
@@ -24,7 +24,6 @@ use Symfony\Component\DependencyInjection\ContainerInterface;
 use Symfony\Component\DependencyInjection\ResettableContainerInterface;
 use Symfony\Component\DomCrawler\Crawler;
 use Symfony\Component\HttpFoundation\Response;
-use Symfony\Component\HttpFoundation\Session\Session;
 use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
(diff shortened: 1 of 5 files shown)


src/Test/WebTestCase.php
```diff
@@ -230,61 +230,42 @@ protected function getContainer(): ContainerInterface
...
      *
-     * @param bool|array $authentication
-     * @param array      $params
+     * @param bool  $authentication
+     * @param array $params
      *
      * @return Client
      */
-    protected function makeClient($authentication = false, array $params = []): Client
+    protected function makeClient(bool $authentication = false, array $params = []): Client
```
alexislefebvre: We may rename it to `makeAuthenticatedClient()` because calling `makeClient()` without using authentication is the same as calling `static::createClient(['environment' => $this->environment], $params);`.

src/Test/WebTestCase.php
```diff
@@ -376,7 +369,12 @@ public function fetchContent(string $path, string $method = 'GET', bool $authent
      */
     public function fetchCrawler(string $path, string $method = 'GET', bool $authentication = false, bool $success = true): Crawler
     {
-        $client = $this->makeClient($authentication);
+        if ($authentication) {
```
lsmith77: in such cases I lean towards ternary syntax, but I am fine keeping it as is.

Issue #507, opened 2019-05-10
Tag 1.9.3 was removed


dominikzogg:
Until yesterday the bundle was installable by the tag 1.9.3 side by side with phpunit 7.5 and worked for our project, now the tag is gone and we're in a situation we cannot recover, cause no newer version is stable and supports phpunit 7.5.

corphi:
The commit 80329104370eabebc10df69c15e98f585773884d was tagged 1.9.3.

lsmith77:
what is the issue with 1.9.5?

dominikzogg:
Conflict with phpunit >= 7

Lukas Kahwe Smith <notifications@github.com> schrieb am Fr., 10. Mai 2019,
11:46:

lsmith77:
@alexislefebvre I guess we could try to get a 1.94.1 one released that disallows Symfony 4 but works for phpunit 7 and Symfony 3. wdyt?

alexislefebvre:
Sorry for the inconvenience.

Does https://packagist.org/packages/liip/functional-test-bundle#1.10.0 fit your needs?

dominikzogg:
@alexislefebvre we're using phpunit 7.5, so no

alexislefebvre:
Sorry, I missed that it was noted in `Conflicts`. :see_no_evil:

alexislefebvre:
I upgraded dependencies in #508, it could be released as `1.11.0` (because requiring PHP 7.1 is a great change, and [1.10.0](https://github.com/liip/LiipFunctionalTestBundle/releases/tag/1.10.0) is the last release from master, not 1.9 :wink:).

corphi:
It would work for me; I am running PHP 7.2 and PHPUnit 7.5, but can’t upgrade to Symfony 3.4 quite yet. I’ll fire up a test run tomorrow. Thanks.

alexislefebvre:
Please try the 1.11. version: https://github.com/liip/LiipFunctionalTestBundle/releases/tag/1.11.0

alexislefebvre:
Please try the 1.11.0 version: https://github.com/liip/LiipFunctionalTestBundle/releases/tag/1.11.0

PR #508, opened 2019-05-10
1.x: Allow PHPUnit 7, remove PHP 5.6 support

alexislefebvre:
See #507

The `void` return type is required for compatibility with PHPUnit 7, so PHP 7.1 is required.


corphi:
It works. :+1:

Issue #509, opened 2019-05-14
Compatibility with Symfony 4.3.0-BETA1


gnutix:
Running any test extending `Liip\FunctionalTestBundle\Test\WebTestCase` result in : 

```
TypeError: Return value of Liip\FunctionalTestBundle\Test\WebTestCase::makeClient()
must be an instance of Symfony\Bundle\FrameworkBundle\Client, instance of
Symfony\Bundle\FrameworkBundle\KernelBrowser returned

/srv/app/vendor/liip/functional-test-bundle/src/Test/WebTestCase.php:413
```

### Preconditions
* Symfony 4.3.0-BETA1
* LiipFunctionalTestBundle 2.0.0-alpha15 and 3.0.0-alpha1

gnutix:
I've fixed the code with this patch : https://diffy.org/diff/ydi0ltz330f1n73szx1xajor

But I'm not sure if I can do a PR, as it's a BC break for this library I guess.

alexislefebvre:
Thanks, could you please open a PR targeting the `3.x` branch? We'll see if it breaks tests.

gnutix:
There you go : https://github.com/liip/LiipFunctionalTestBundle/pull/510

alexislefebvre:
Thanks for the PR, it fixed the issue for Symfony 4.3 but break tests with all other versions.

I'm still wondering if this is an issue from Symfony that require a change to avoid BC break, see these related issues: https://github.com/symfony/symfony/pull/30541 https://github.com/symfony/symfony/issues/31039 https://github.com/symfony/symfony/pull/31040

alexislefebvre:
We could use an alias as shown in this comment: https://github.com/symfony/symfony/pull/30541#issuecomment-481215018

gnutix:
Not sure how. I get `PHP Fatal error:  Uncaught Cannot declare class Symfony\Component\BrowserKit\Client, because the name is already in use` if I put `class_alias(\Symfony\Component\BrowserKit\AbstractBrowser::class, 'Symfony\Component\BrowserKit\Client');` on top of my `WebTestCase` class (that extends this bundle's one and override `makeClient`).

gnutix:
PR ready.

alexislefebvre:
Thanks a lot! I'll publish a new release.

PR #510, opened 2019-05-14
Replace Client by KernelBrowser for Symfony 4.3 compatibility.

gnutix:
See #509


alexislefebvre:
You can add Symfony 4.3 in the matrix of `.travis.yml`.

In the issue you wrote:

> TypeError: Return value of Liip\FunctionalTestBundle\Test\WebTestCase::makeClient()
must be an instance of Symfony\Bundle\FrameworkBundle\Client, instance of
Symfony\Bundle\FrameworkBundle\KernelBrowser returned

I don't understand why we must return `KernelBrowser` when the error stats that it must be an instance of `Client`. :thinking:

gnutix:
> I don't understand why with this PR we return KernelBrowser when the error states that it must be an instance of Client.

The error is that LiipFunctionalTestBundle's WebTestCase defines a strict return type of the "Client" class, but Symfony 3.4 replaced it by "KernelBrowser" (which is basically empty and simply extends the Client). So PHP doesn't like that Symfony's code return an instance of "KernelBrowser" and LiipFunctionalTestBundle's WebTestCase try to return a "Client".

gnutix:
@alexislefebvre Any idea how I can set the `"minimum-stability": "beta"` only for one test matrix? There's no composer flag for that AFAIK.

alexislefebvre:
You can require `^4.3@alpha`, it should work: https://semver.mwl.be/#?package=symfony%2Fsymfony&version=%5E4.3&minimum-stability=alpha

gnutix:
The tests are passing with Symfony 4.3.0-BETA1 (and failing everywhere else). Anything else I should do?

alexislefebvre:
Thanks for the PR.

We should try to use a compatibility layer instead of having a version of the bundle that require Symfony 4.3.

See my comments here: https://github.com/liip/LiipFunctionalTestBundle/issues/509#issuecomment-494739539

gnutix:
Replaced by https://github.com/liip/LiipFunctionalTestBundle/pull/512

PR #545, opened 2019-08-26
Command Inputs

rodnaph:
**Is your feature request related to a problem? Please describe.**
I'm writing a test for a command which accepts user input, so need access to configure the `CommandTester` inputs.

**Describe the solution you'd like**
I'd like to be able to easily configure the inputs to the command I am testing.

**Describe alternatives you've considered**
I considered splitting the current `runCommand` into two parts, one for building the command (which would create a `CommandTester` instance and return it, where it could additionally configured by specific tests) and another for running it. This felt awkward though because the configure/run parameters are better viewed in one place.

**Additional context**
Please let me know if there are better ways to approach this that are already supported, I couldn't find anything in the issues that referenced configuring user input.

### DIFF
--- a/src/Test/WebTestCase.php
+++ b/src/Test/WebTestCase.php
@@ -49,6 +49,11 @@ abstract class WebTestCase extends BaseWebTestCase
     protected $decorated;
 
     /**
+     * @var array|null
+     */
+    private $inputs = null;
+
+    /**
      * @var array
      */
     private $firewallLogins = [];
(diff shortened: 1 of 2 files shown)


alexislefebvre:
Thanks for the PR.

The `liipfunctionaltestbundle:test` command doesn't accept input. The `testRunCommandWithInputs` doesn't check if the command succeeded. It may be interesting to add an optional input in the command and check in the test that it works, for example by outputting the input. So we could be sure that inputs are passed properly.

Could you please do this?

alexislefebvre:
I added tests in #546.

rodnaph:
Thanks Alex, sorry I couldn’t get to it first.

PR #550, opened 2019-11-23
Replaced deprecated assertContains with assertStringContainsString

spajxo:
Hi, just replaced deprecated PHPUnit assert so that tests won't throw warnings.

#SymfonyHackday


alexislefebvre:
Thanks!

PR #567, opened 2020-05-15
Create WebTestCase::loginClient to allow working with a single kernel

ndench:
Booting a kernel before calling Symfony WebTestCase::createClient() is deprecated in Symfony 4.4 (symfony/symfony#32056). This exposed an issue with LiipTestFixturesBundle which was causing multiple kernels to be booted during a test. After fixing this in liip/LiipTestFixturesBundle#62 it was brought to my attention that Liip WebTestCase::loginAs is now broken since we require the client to be created before loading fixtures.

As a result, I think deprecating the `loginAs` method and adding a `loginClient` method is the best path forward. Note that in Symfony 5.1 we have  [`$client->loginUser()`](https://symfony.com/blog/new-in-symfony-5-1-simpler-login-in-tests) which is very similar.

I've been using a similar helper function in my projects for a couple of years now, and after suggesting it as an option on https://github.com/liip/LiipTestFixturesBundle/pull/62#issuecomment-622191412 others have reported that it works for them as well.

### DIFF
--- a/doc/logged.md
+++ b/doc/logged.md
@@ -8,7 +8,7 @@ You have three alternatives to create an already logged in client:
 
 1. Use the `liip_functional_test.authentication` key in the `config_test.yml` file;
 2. Pass an array with login parameters directly when you call the method;
-3. Use the method `WebTestCase::loginAs()`;
+3. Use the method `WebTestCase::loginClient()`;
 
 ### Logging in a user from the `config_test.yml` file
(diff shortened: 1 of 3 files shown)


alexislefebvre:
Thanks! You will find it in last version: https://github.com/liip/LiipFunctionalTestBundle/releases/tag/4.2.0

PR #568, opened 2020-05-22
[Doc] add doctrine profiling part to enable query counter

ottaviano:
need to enable profiling on doctrine bundle too otherwise, the profiler will not record the queries.


alexislefebvre:
Sorry for the delay. Thanks!

PR #574, opened 2020-10-22
Update twig to 3

jschlaht:
Trying to use twig in version 3 with this project.


jschlaht:
Unfortunately I can not see, what exactly was the problem in scrutinizer. I run tests locally with PHPUnit version 9 and 8 and with twig in version 2 and 3, all without problems.

alexislefebvre:
Thanks!

PR #582, opened 2021-03-12
WebTestCase memory leak fix

shtumi:
There is memory leak in WebTestCase because method getContainer creates new kernel which is never shutdown.
So in my project I was need more then 4Gb of memory to pass about 800 tests.
After fix I need less then 300Mb.

This problem was discussed here:
https://github.com/symfony/symfony/issues/35164

And also it was fixed in similar (but not exact) way inside LiipTestFixturesBundle:
https://github.com/liip/LiipTestFixturesBundle/pull/62

### DIFF
--- a/src/Test/WebTestCase.php
+++ b/src/Test/WebTestCase.php
@@ -25,6 +25,7 @@ use Symfony\Component\DependencyInjection\ContainerInterface;
 use Symfony\Component\DependencyInjection\ResettableContainerInterface;
 use Symfony\Component\DomCrawler\Crawler;
 use Symfony\Component\HttpFoundation\Response;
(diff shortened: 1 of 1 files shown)


alexislefebvre:
Thanks! Please fix the issue reported by StyleCI.

alexislefebvre:
@shtumi could you please rebase the PR and fix the [error reported by StyleCI](https://github.styleci.io/analyses/7anD3j)?

shtumi:
really currently code of WebTestCase is changed a lot and I'm really not sure that problem is still exists and if exists, then then it's more easy to take new fork, investigate the problem and give new solution.

PR #584, opened 2021-06-01
Fix new static method getContainer in KernelTestCase collision

k20human:
Proposition to fix issue #583

PR #585, opened 2021-06-01
Run tests with Symfony 5.3

alexislefebvre:


alexislefebvre:
It should be fixed in #584

PR #587, opened 2021-07-02
Fix Symfony 5.2 deprecation

jmclean:
Lately in my projects I've been getting this deprecation warning when running tests:

> Since symfony/framework-bundle 5.2: Accessing the "liip_functional_test.validator" service directly from the container is deprecated, use dependency injection instead.

That service has `public="true"` so I'm not sure why this warning happens. Maybe it's some nuance of service decoration.

This can be fixed by going through `test.service_container`.


alexislefebvre:
Thanks!

alexislefebvre:
Here is a release with this patch: https://github.com/liip/LiipFunctionalTestBundle/releases/tag/4.4.1

PR #588, opened 2021-08-03
Replaces call to deprecated isMasterRequest with isMainRequest

bmupex:

PR #591, opened 2021-08-20
Fix issue with isMainRequest

### LINKED ISSUE #590: Call to deprecated method isMasterRequest() when isMainRequest() returns false
Relevant piece of code in `\Liip\FunctionalTestBundle\Validator\DataCollectingValidator`
```
    public function onKernelRequest(GetResponseEvent $event): void
    {
        if (method_exists($event, 'isMainRequest') && $event->isMainRequest()) {
            $this->clearLastErrors();
        } elseif ($event->isMasterRequest()) {
            $this->clearLastErrors();
        }
    }
```

The method `onKernelRequest()` correctly uses `$event->isMainRequest()` if the method exists, however the way the condition is constructed, the elseif condition is checked when `isMainRequest()` returns false. As a result, the deprecated `isMasterRequest()` method is called unnecessarily.

### Preconditions
1. Using composer package `liip/functional-test-bundle  4.4.2`

### Steps to reproduce
The bug is obvious when looking at the code. To reproduce, call 
`\Liip\FunctionalTestBundle\Validator\DataCollectingValidator::onKernelRequest()` with `\Symfony\Component\HttpKernel\HttpKernelInterface::SUB_REQUEST` as the last parameter

### Expected result
1. A deprecation error is not triggered when `isMainRequest()` exists and returns false.

### Actual result
1. When `isMainRequest()` exists and returns false, a deprecation error is triggered: `php.INFO: User Deprecated: Since symfony/http-kernel 5.3: "Symfony\Component\HttpKernel\Event\KernelEvent::isMasterRequest()" is deprecated, use "isMainRequest()" instead.`

alexislefebvre:
Fixes #590

See #588

### DIFF
--- a/src/Validator/DataCollectingValidator.php
+++ b/src/Validator/DataCollectingValidator.php
@@ -88,9 +88,17 @@ class DataCollectingValidator implements ValidatorInterface, EventSubscriberInte
 
     public function onKernelRequest(GetResponseEvent $event): void
     {
-        if (method_exists($event, 'isMainRequest') && $event->isMainRequest()) {
-            $this->clearLastErrors();
-        } elseif ($event->isMasterRequest()) {
+        // The isMainRequest method has been added in Symfony 5.3
+        if (method_exists($event, 'isMainRequest')) {
+            if ($event->isMainRequest()) {
+                $this->clearLastErrors();
+            }
+
+            return;
+        }
+
+        // For Symfony < 5.3, call the legacy method
+        if ($event->isMasterRequest()) {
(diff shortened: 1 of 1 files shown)


tperunsky (approved):
Looks good to me. In my environment it fixes the issue.

alexislefebvre:
@tperunsky Thanks for the test, the fix is available in the last release: https://github.com/liip/LiipFunctionalTestBundle/releases/tag/4.4.3

tperunsky:
Thank you for the quick fix!

PR #593, opened 2021-12-09
Added a fix for symfony 5.3 container.

### LINKED ISSUE #592: Regression in getContainer with symfony 5.3
After the fix #584 there is no possibility to test `runCommand` with `$reuseKernel = true` when using symfony 5.3+.

### Preconditions
1. Use symfony 5.4 
2. Use `runCommand` test.

When there is a `$this->getContainer()` call before the `runCommand` it's not possible to reuse same kernel.
In the following code there is no possibility to reuse kernel created by `getContainer`.

```php
$client = $this->createMock(Client::class);

$this->getContainer()->get(Service::class)->setClient($client);
$this->runCommand('command', [], true);
```

biozshock:
Fixes #592

### DIFF
--- a/src/Test/WebTestCase.php
+++ b/src/Test/WebTestCase.php
@@ -25,6 +25,7 @@ use Symfony\Component\DependencyInjection\ContainerInterface;
 use Symfony\Component\DependencyInjection\ResettableContainerInterface;
 use Symfony\Component\DomCrawler\Crawler;
 use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\KernelInterface;
 use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 use Symfony\Component\Security\Core\Authentication\Token\TokenInterface;
 use Symfony\Component\Security\Core\Authentication\Token\UsernamePasswordToken;
@@ -40,10 +41,11 @@ if (!class_exists(Client::class)) {
  * @author Benjamin Eberlei <kontakt@beberlei.de>
  *
  * @method ContainerInterface getContainer()
+ * @property string $environment;
  */
 abstract class WebTestCase extends BaseWebTestCase
 {
-    protected $environment = 'test';
+    protected static $env = 'test';
 
     protected $containers;
 
@@ -70,7 +72,7 @@ abstract class WebTestCase extends BaseWebTestCase
      */
     protected function getServiceMockBuilder(string $id): MockBuilder
     {
-        $service = $this->getDependencyInjectionContainer()->get($id);
+        $service = $this->getContainer()->get($id);
         $class = get_class($service);
 
         return $this->getMockBuilder($class)->disableOriginalConstructor();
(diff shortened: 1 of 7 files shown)


biozshock:
No, that's the code needed to properly boot a kernel with the selected environment with static function `createKernel`

alexislefebvre:
Why `$environment` was removed?

biozshock:
it's not static.

alexislefebvre:
Ok, thank you.

So we can't keep `$environment` because it can't be accessed in `createKernel()` ?

Please add a note in https://github.com/liip/LiipFunctionalTestBundle/blob/master/CHANGELOG.md to explain how to upgrade. It looks like it's a BC break so it may require a new major release.

biozshock:
That's not a BC break, because you may still use old `environment` property without any consequences, except you'd get a silenced deprecation message.
Next major release should remove the deprecation and `environment` property.

src/Test/WebTestCase.php
```diff
@@ -224,22 +226,60 @@ protected function getDependencyInjectionContainer(): ContainerInterface
...
             return $this->getDependencyInjectionContainer();
         }
 
         throw new \Exception("Method {$name} is not supported.");
     }
 
+    public function __set($name, $value)
+    {
+        if ($name !== 'environment') {
+            throw new \Exception(sprintf('There is no property with name "%s"', $name));
```
alexislefebvre, line 253: What if I extend WebTestCase and I use `$this->client` (or whatever new variable), will it throw an exception?

tests/Command/CommandTest.php
```diff
@@ -85,8 +85,10 @@ public function testRunCommandWithoutOptionsAndNotReuseKernel(): void
         $this->assertIsBool($this->getDecorated());
         $this->assertTrue($this->getDecorated());
 
-        // Run command and not reuse kernel
-        $this->environment = 'prod';
```
alexislefebvre, line 89: Please duplicate this test to use `$this->environment = 'prod';` and not `self::$env = 'prod';`, to ensure BC.

src/Test/WebTestCase.php
alexislefebvre: Actually some tests already declare new variables (e.g. [`$this->client =`](https://github.com/liip/LiipFunctionalTestBundle/pull/593/files#diff-2a2f8a611efd3fe4ee275790ac720835aab87fdcef0ba731d3754e40a3b92f01)) but I don't understand why it doesn't trigger this exception.

biozshock: This code is executed only when the variable does not exist. See https://www.php.net/manual/en/language.oop5.overloading.php#object.set

alexislefebvre:
Could you please fix the failing test? Do you need help?

tests/Command/CommandTest.php
biozshock: done

alexislefebvre:
Thank you! These changes are available in the last release: https://github.com/liip/LiipFunctionalTestBundle/releases/tag/4.5.0

mikysan:
> That's not a BC break, because you may still use old environment property without any consequences, except you'd get a silenced deprecation message.
> Next major release should remove the deprecation and environment property.

This is actually a BC break, in my company test suite the `WebTestCase` class was extended and `$environment` protected property overwritten, updating this library broke all the functional tests because the environment was once again defaulted to "test" (without us knowing it). Also the tests errors were about missing configuration without the environment explicitly showed by phpunit and no deprecation warning raised.

Also note that overriding `protected static $env = 'test';` in the extend class  doesn't fix the issue due to the use of `self::$env`  instead of `static::$env`. So there are actually two issues...

I didn't manage to find any issue about this, should I open them? If there's one already opened could you please link it?

alexislefebvre:
Thanks for the report and sorry for the inconvenience.

There's no issue about this regression yet, feel free to add an issue and/or open a PR.

We could restore `protected $environment = 'test';` and a method to read `self::$env`, or `$this->environment` if the former has the default value?

mikysan:
Ok, I'll open 2 issues and details there. Btw I won't be able to contribute with any PR until Friday hope it's ok.

Issue #595, opened 2022-01-17
Can't overwrite WebTestCase $env static property


mikysan:
### Preconditions

1. PHP 8.0
2. Symfony 5.4
3. liip/functional-test-bundle 4.5

### Steps to reproduce

1. rename Symfony test environment to testing
2. create a test like following
```
namespace App\Tests;

use Liip\FunctionalTestBundle\Test\WebTestCase;

class FunctionalTestCase extends WebTestCase
{
    protected static $env = 'testitng';

    public function testFoo()
    {
        $this->assertTrue(true);
    }
}
```

### Expected result
The test should pass

### Actual result
test fail with `You have requested a non-existent parameter "secret".` error message.

### Proposed solution
Replace `self::$env` with `static::$env` in related files (see: https://github.com/liip/LiipFunctionalTestBundle/pull/593/files)

PR #596, opened 2022-01-21
Fix access to static $env property  form 'self::$env' to 'static::$env'

mikysan:
This fix allows to extend `WebTestCase` and override the static `$env` property value effectively.

Closes issue #595


alexislefebvre:
Thanks! Please try the last release: https://github.com/liip/LiipFunctionalTestBundle/releases/tag/4.5.1

PR #598, opened 2022-01-27
[WIP] Sf 6 compatibility

magnetik:
Trying to make this bundle compatible with Symfony 6

ref #597

PR #603, opened 2022-05-13
Support Symfony 6

alexislefebvre:
Based on #598

`loginUser()` is called when it's available, it was added in Symfony 5.1: https://symfony.com/doc/5.4/testing.html#logging-in-users-authentication

Without it, tests were broken. But with it, it may break some tests suites.


CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Changelog
 
+## 5.0
+
+## Removed 
+
+## 4.6
+
+Deprecated […] methods, use loginUser() from Symfony instead https://symfony.com/doc/5.4/testing.html#logging-in-users-authentication
```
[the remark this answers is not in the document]
alexislefebvre: Thanks for your input!

`WebTestCase::isSuccessful` could be replaced by [`assertResponseIsSuccessful`](https://symfony.com/doc/4.4/testing.html#response-assertions) which exist since Symfony 4.4.

`DatabaseToolCollection` comes from the other bundle: https://github.com/liip/LiipTestFixturesBundle, which should work with Symfony 6.6.

tests/App/session_4.yml
```diff
@@ -0,0 +1,7 @@
+framework:
```
alexislefebvre, line 1: use `include` to put the parameters compatible with all versions

alexislefebvre: It's `imports`, not `include`: https://github.com/liip/LiipFunctionalTestBundle/commit/122d254c733b5cde2a29f566369cf23cc2283875

kl3sk:
With a simple test this error show:
```bash
Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException: You have requested a non-existent service "Liip\TestFixturesBundle\Services\DatabaseToolCollection".
```

Used like this in `class UserTest extends KernelTestCase` :
```php
    public function setUp(): void
    {
        parent::setUp();

        $this->databaseTool  = static::getContainer()->get(DatabaseToolCollection::class)->get();
        $this->entityManager = static::getContainer()->get(EntityManagerInterface::class);

        $this->users = $this->entityManager->getRepository(User::class)->findAll();
    }
```

As current doc said:
in `config/test/framework.yaml`

```yaml
framework:
    test: true
    session:
        storage_id: session.storage.mock_file
        name: MOCKSESSION

liip_functional_test: ~
```

alexislefebvre:
@kl3sk Thanks, I'll add it to the doc of the other bundle: https://github.com/liip/LiipTestFixturesBundle/pull/206

kl3sk:
Ok thanks,

Just to be clear. This doen't work for me.

alexislefebvre:
Thanks, I'll add this too for Symfony 6.

kl3sk:
This is a good question @alexislefebvre 🤣 

I have a default installation and don't understand why this happen.
A though because the compatibility is alpha at least.

PR #613, opened 2023-03-30
allow doctrine/annotations 2

alexislefebvre:
Fixes #612

PR #614, opened 2023-04-04
composer: move doctrine/annotations to require-dev and suggest

alexislefebvre:
It may be done after #613

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -64,6 +64,8 @@ jobs:
           symfony/validator=${{ matrix.symfony-version }}
           symfony/yaml=${{ matrix.symfony-version }}
 
+      # This is needed to fix builds where the `annotation_reader` service may not be set up if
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -65,6 +65,11 @@ jobs:
           symfony/validator=${{ matrix.symfony-version }}
           symfony/yaml=${{ matrix.symfony-version }}
 
+      # This is needed to fix builds where the `annotation_reader` service may not be set up if
+      # the Annotations package is not in the production dependencies
+      - name: Move Annotations to require
+        run: composer require --no-update "doctrine/annotations:^1.8.0|^2.0"
```
alexislefebvre, line 71: Taken from https://github.com/liip/LiipTestFixturesBundle/commit/6acd4cc8fe23752c8609b64db24dc1fd8dd80a8f

Issue #630, opened 2024-01-04
Symfony 7 compatibility


kl3sk:
**Is your feature request related to a problem? Please describe.**
No upgrade to SF 7 possible

**Describe the solution you'd like**
Upgrade to SF 7

**Describe alternatives you've considered**
Downgrade to SF 6.4

Thanks

alexislefebvre:
Thanks for the report.

We can take inspiration from:

- https://github.com/liip/LiipTestFixturesBundle/pull/248

alexislefebvre:
Free free to open a Pull Request with the required changes.

kl3sk:
Ok why not, I'll have a look.

alexislefebvre:
Could you please tell us what methods of this bundle are used in your projects? Some methods have been integrated in Symfony.

kl3sk:
Crap ! You nailed it, none seems to be used any more.

I'll remove the bundle from my project and try some tests that pass 🙄

kl3sk:
Because I was writing functional tests I though this bundle was used,

But it is `LiipTestFixturesBundle` that is used. And seems already ready for Sf7.

My bad.

Nirator78:
Hi,
I'm currently using `LiipFunctionalTestBundle` in my Symfony `6.4` and i can't update in Symfony `7.0` because `LiipFunctionalTestBundle` is not compatible.

> Your requirements could not be resolved to an installable set of packages.
 >  Problem 1
 >    - Root composer.json requires liip/functional-test-bundle 4.10.0 -> satisfiable by liip/functional-test-bundle[4.10.0].
>     - liip/functional-test-bundle 4.10.0 requires symfony/browser-kit ^4.4 || ^5.1 || ^6.0 -> found symfony/browser-kit[v4.4.0, ..., v4.4.44, v5.1.0, ..., v5.4.31, v6.0.0, ..., v6.4.0] but it conflicts with your root composer.json require (7.0.*).

alexislefebvre:
@Nirator78 Could you please tell us what methods of this bundle are used in your projects? Some methods have been integrated in Symfony.

See also:

- GH-633

alexislefebvre:
Please try the last release: https://github.com/liip/LiipFunctionalTestBundle/releases/tag/4.11.0

Nirator78:
I'm actually using the makeClient method and the assertStatusCode method.

alexislefebvre:
[...]

PR #631, opened 2024-01-05
Symfony 7 compatibillity

kl3sk:
This may allow Symfony 7 support.

Older version were dropped and older PHP version.

Fell free to comment/annotate 😃 

Fixes #630

PR #634, opened 2024-01-07
chore: allow Symfony 7

alexislefebvre:
Fixes GH-630

Based on:

- GH-631


alexislefebvre:
We have this error with Symfony 7:

> Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException: The service "liip_functional_test.query.counter" has a dependency on a non-existent service "annotation_reader".

I don't want to add a compatibility layer to support `QueryCount` as an annotation and an attribute. :grimacing:

PR #670, opened 2025-12-11
feat: split tests of `query.max_query_count`

### LINKED ISSUE #606: Restore test with max_query_count: 1
One test of `AllowedQueriesExceededException` and some other things were broken on the main branch, I fixed these issues in another PR: #604 I'm not happy with the change of `max_query_count` from `1` to `0` in the internal tests but if it allow to have a version compatible with Symfony 6, it will be worth it.

_Originally posted by @alexislefebvre in https://github.com/liip/LiipFunctionalTestBundle/issues/598#issuecomment-1126584504_

---

Here is the related commit: https://github.com/liip/LiipFunctionalTestBundle/pull/604/commits/93a080a3adf9dfef6cc434c50fc092c90cca42c9

alexislefebvre:
Fixes GH-606

- GH-606

Split from:

- GH-668

TODO:

- [ ] update `QueryCounter` so that it doesn’t fail if the annotation reader is not available, `query.max_query_count` should work on Symfony 7+ even without the support for annotations. Show a warning instead if the annotation can’t be read? The bundle need to the annotation reader to check if an `@QueryCount(x)` annotation was actually used…
- [ ] consider adding an attribute `QueryCount` to replace the annotation on the long term

Update:

`testAllowedQueriesExceededException` worked because of `query.max_query_count: 0`, not because of the annotation. It was a false positive since e2fdc70cd1f7a9af6759d5fd9406f16bda4ac7f2 from

- GH-604

### DIFF
--- a/src/QueryCounter.php
+++ b/src/QueryCounter.php
@@ -16,6 +16,7 @@ namespace Liip\FunctionalTestBundle;
 use Doctrine\Common\Annotations\Reader;
 use Liip\FunctionalTestBundle\Annotations\QueryCount;
 use Liip\FunctionalTestBundle\Exception\AllowedQueriesExceededException;
+use Symfony\Component\HttpKernel\Kernel;
 
 final class QueryCounter
 {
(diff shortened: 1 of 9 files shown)


src/QueryCounter.php
```diff
@@ -50,8 +51,16 @@ private function getMaxQueryCount(): ?int
 
     private function getMaxQueryAnnotation(): ?int
     {
+        if (Kernel::MAJOR_VERSION >= 7) {
+            trigger_error('The annotationReader is not available and it can’t be enabled on Symfony 7+, @QueryCount is not supported', \E_USER_WARNING);
```
alexisLefebvre, line 55: Providing this information is better than before, but this warning may not appear when running tests, so it may be ineffective.

